PR #3, opened 2016-10-28
WIP: Validation

jml:
Depends on #2

---

This change is [Reviewable](https://reviewable.io/reviews/jml/graphql-api/3)


src/GraphQL/Output.hs
```diff
@@ -0,0 +1,76 @@
...
+--   * be a map
+--   * have a "data" key iff the operation executed
+--   * have an "errors" key iff the operation encountered errors
+--   * not include "data" if operation failed before execution (e.g. syntax errors,
+--     validation errors, missing info)
+--   * not have keys other than "data", "errors", and "extensions"
+--
+-- Other interesting things:
+--
+--   * Doesn't have to be JSON, but does have to have maps, strings, lists,
```
teh: interestingly it's almost like a protobuf - and I'm sure we could encode it as that (but for internal RPC sth like a `Binary` encoding might be faster)

src/GraphQL/Output.hs
```diff
@@ -0,0 +1,76 @@
...
+--
+--   * must be a non-empty list
+--   * each error is a map with "message", optionally "locations" key
+--     with list of locations
+--   * locations are maps with 1-indexed "line" and "column" keys.
+data Response
+  = Success Map
+  | PreExecutionFailure Errors
+  | ExecutionFailure Errors
+  | PartialSuccess Map Errors
```
teh: `PartialSuccess` doesn't make sense if errors/data is either/or (i.e. an `Either`) as described above

src/GraphQL/Validation.hs
```diff
@@ -0,0 +1,103 @@
...
+union HumanOrAlien = Human | Alien
+
+type QueryRoot {
+  dog: Dog
+}
+-}
+
+newtype ValidDocument = Valid AST.Document deriving (Eq, Show)
+
+validate :: Alternative m => AST.Document -> m ValidDocument
```
teh: what's the purpose of validate? Make sure the query is valid? Matches the types?

src/GraphQL/Value.hs
```diff
@@ -0,0 +1,156 @@
...
+fieldSetToMap (FieldSet fields) = Map (Map.fromList [ (name, value) | Field name value <- toList fields ])
+
+empty :: FieldSet
+empty = FieldSet Set.empty
+
+singleton :: Field -> FieldSet
+singleton = FieldSet . Set.singleton
+
+-- TODO: Fail on duplicate keys.
+union :: Alternative m => FieldSet -> FieldSet -> m FieldSet
```
teh: is `union` related to graphql union types? If so I'm not sure why it should fail on duplicates

src/GraphQL/Output.hs
jml: Yep. Or thrift, which is I think what FB use.

src/GraphQL/Output.hs
jml: I'm pretty sure it does make sense (although maybe my comment doesn't). 

> If the data entry in the response is not null, the errors entry in the response may contain any errors that occurred during execution. If errors occurred during execution, it should contain those errors.

From https://facebook.github.io/graphql/#sec-Errors

Note that `data` can be absent or null, and there are different meanings associated with each.

jml:
I'll wait until I've had a chance to review your work and talk to you about it before I merge this.

teh:
Feel free to merge - I think this is good as is

PR #5, opened 2016-12-12
Type api

teh:
This will be the "engineering" branch, i.e. by the time we merge I want it to be in good enough shape to be usable in a library.


jml:
> This will be the "engineering" branch, i.e. by the time we merge I want it to be in good enough shape to be usable in a library.

FWIW, I'm not sure this is the best approach. The whole repo is pre-release. We might as well take this work, do the minimum possible such that:
- it compiles
- the tests pass
- gaps & defects are clearly marked with comments or warning-generating code

Then merge it to master.

In other words, master should be the point of collaboration.

teh:
OK, makes sense. I'll fix the last few big items (like `error` remaining in code) and then we cen reviiew & merge

PR #9, opened 2016-12-13
Mopup 1

teh:


src/GraphQL/TypeApi.hs
```diff
@@ -147,18 +152,22 @@ instance forall ks t m. (KnownSymbol ks, HasGraph m t, MonadThrow m, MonadIO m)
...
-      Right v' -> buildFieldResolver @m @f (handler v') selection
-  buildFieldResolver _ f = ("", queryError ("buildFieldResolver got non AST.Field" <> show f <> ", query probably not normalized"))
+    in case v of
+       Nothing -> case valueMissing @t argName of
+                     Left err' -> ("", queryError err')
+                     Right v' -> buildFieldResolver @m @f (handler v') selection
+       Just v' -> case readValue @t v' of
+                    Left err' -> ("", queryError err')
+                    Right v'' -> buildFieldResolver @m @f (handler v'') selection
+  buildFieldResolver _ f = ("y", queryError ("buildFieldResolver got non AST.Field" <> show f <> ", query probably not normalized"))
```
jml: You should probably remove the nested case by doing:

```haskell
v' = maybe (valueMissing @t argName) (readValue @t)
in case v' of
  Left err' -> ("", queryError ')
  Right v'' -> buildFieldResolver @m @f (handler v'') selection
```

tests/TypeTests.hs
```diff
@@ -11,8 +11,15 @@ import GraphQL.Schema
...
+data DogCommandEnum = Sit | Down | Heel
+
+instance GraphQLEnum DogCommandEnum where
+  enumValues = ["SIT", "DOWN", "HEEL"]
+  enumToValue _ = undefined
+  enumFromValue _ = undefined
+
 -- Alternative might be a sum type with deriving Generic and 0-arity constructors?
-type DogCommand = Enum "DogCommand" '["SIT", "DOWN", "HEEL"]
+type DogCommand = Enum "DogCommand" DogCommandEnum
```
jml: This is an unfortunate change ergonomically, but I can see why it might be beneficial for us figuring out a good internal model. No change required.

src/GraphQL/TypeApi.hs
```diff
@@ -147,18 +152,22 @@ instance forall ks t m. (KnownSymbol ks, HasGraph m t, MonadThrow m, MonadIO m)
     let childResolver = buildResolver @m @t handler selectionSet
         name = toS (symbolVal (Proxy :: Proxy ks))
     in (name, childResolver)
-  buildFieldResolver _ f = ("", queryError ("buildFieldResolver got non AST.Field" <> show f <> ", query probably not normalized"))
+  buildFieldResolver _ f = ("x", queryError ("buildFieldResolver got non AST.Field" <> show f <> ", query probably not normalized"))
```
jml: Not for this PR.

It's not clear to me what this tuple represents. Perhaps it should be a real data type.

src/GraphQL/TypeApi.hs
```diff
@@ -87,14 +93,18 @@ instance forall m hg. (MonadThrow m, MonadIO m, HasGraph m hg) => HasGraph m (Li
...
 -- each argument in the list) but considering the graphql use case
 -- where N usually < 10 this is probably OK.
--- TODO (Maybe Int) types that can convert Nothing
-lookupValue :: AST.Name -> [AST.Argument] -> Either Text AST.Value
+lookupValue :: AST.Name -> [AST.Argument] -> Maybe AST.Value
 lookupValue name args = case find (\(AST.Argument name' _) -> name' == name) args of
-  Nothing -> Left ("Argument not found:" <> name)
-  Just (AST.Argument _ value) -> pure value
+  Nothing -> Nothing
+  Just (AST.Argument _ value) -> Just value
```
jml: Optional

Define and/or find helpers for:

```haskell
argumentName (AST.Argument name _) = name
argumentValue (AST.Argument _ value) = value
```

And write this as:

```haskell
argumentValue <$> find ((name ==) . argumentName) args
```

src/GraphQL/TypeApi.hs
```diff
@@ -50,17 +50,23 @@ queryError = throwM . QueryError
...
+-- | The ReadValue instance converts AST.Value types like ValueInt to
+-- the type expected by the handler function. It's the boundary
+-- between incoming types and your custom application Haskell types.
 class ReadValue a where
+  -- | Convert the already-parsed value into the type needed for the
+  -- function call.
   readValue :: AST.Value -> Either Text a
+
+  -- | valueMissing is a separate function so we can provide default
+  -- values for certain cases. E.g. there is an instance for `Maybe a`
```
jml: Haddock for "fixed width" is `@@Maybe a@@`. Awesome, huh?

src/GraphQL/TypeApi.hs
```diff
@@ -50,17 +50,23 @@ queryError = throwM . QueryError
...
   type HandlerType m a
-  buildResolver :: HandlerType m a -> AST.SelectionSet -> m GValue.Value
+  buildResolver :: HandlerType m a -> CanonicalQuery -> m GValue.Value
 
--- Parse a value of the right type from an argument
--- TODO
+-- | The ReadValue instance converts AST.Value types like ValueInt to
+-- the type expected by the handler function. It's the boundary
+-- between incoming types and your custom application Haskell types.
 class ReadValue a where
```
jml: Not for this PR.

I would like us to change this to think a bit about terminology before doing a publicly announced release.

In particular, I think we should use `ToValue` and `FromValue` (or some variation thereof) to parallel aeson.

src/GraphQL/Definitions.hs
```diff
@@ -60,17 +60,13 @@ instance forall a as. (HasFieldDefinition a, HasFieldDefinitions as) => HasField
...
-
-instance GetSymbolList '[] where
-  getSymbolList = []
-
+-- | For each enum type we need 1) a list of all possible values 2) a
+-- way to serialise and 3) deserialise.
+class GraphQLEnum a where
+  enumValues :: [Text] -- todo maybe use GValue.Value instead of Text?
+  enumFromValue :: GValue.Value -> Either Text a
+  enumToValue :: a -> GValue.Value
```
jml: I'm not a huge fan of these names, but can't articulate why and think of anything better at the moment. As mentioned in another comment, we should do a naming / vocabulary review before releasing.

src/GraphQL/Definitions.hs
```diff
@@ -60,17 +60,13 @@ instance forall a as. (HasFieldDefinition a, HasFieldDefinitions as) => HasField
...
-instance forall a as. (KnownSymbol a, GetSymbolList as) => GetSymbolList (a:as) where
-  getSymbolList = (toS (symbolVal (Proxy :: Proxy a))):(getSymbolList @as)
-
-instance GetSymbolList '[] where
-  getSymbolList = []
-
+-- | For each enum type we need 1) a list of all possible values 2) a
+-- way to serialise and 3) deserialise.
+class GraphQLEnum a where
+  enumValues :: [Text] -- todo maybe use GValue.Value instead of Text?
```
jml: Or perhaps `Name` (from `GraphQL.Value`)

src/GraphQL/Definitions.hs
```diff
@@ -60,17 +60,13 @@ instance forall a as. (HasFieldDefinition a, HasFieldDefinitions as) => HasField
...
-
-instance forall a as. (KnownSymbol a, GetSymbolList as) => GetSymbolList (a:as) where
-  getSymbolList = (toS (symbolVal (Proxy :: Proxy a))):(getSymbolList @as)
-
-instance GetSymbolList '[] where
-  getSymbolList = []
-
+-- | For each enum type we need 1) a list of all possible values 2) a
+-- way to serialise and 3) deserialise.
+class GraphQLEnum a where
```
jml: Let's investigate this in a different PR to keep momentum going.

Are you aware that Haskell has a built-in [`Enum`](https://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Enum) typeclass? I wonder if we could benefit from that somehow.

src/GraphQL/TypeApi.hs
teh: +1 on naming. I we have a slightly different API with an extra "missingValue" function which would be useful for default values. If you have bandwidth please change naming as you see fit.

src/GraphQL/TypeApi.hs
teh: updated

src/GraphQL/TypeApi.hs
teh: classic lens territory but I didn't feel like pulling in lens for this API. Without lens I find the local pattern match more readable (no indirection through really small helpers)

src/GraphQL/TypeApi.hs
teh: +1, will fix in next branch

src/GraphQL/TypeApi.hs
teh: nice, fixed

tests/TypeTests.hs
teh: The ergonomics of the API definition are worse, but when you have to use the Enum in your own code I think it's better overall. I've been on the fence here for a while but after the Git and Filesystem examples I'm pretty sure that we want the type-class enum (+ DeriveGeneric)

src/GraphQL/Definitions.hs
teh: Yep, `Enum` type class is not what we want. That's used to do e.g. [Dog..Cat] (like [1..10]), max and min bounds. What we need from the enum are the three things I mentioned in the comment.

src/GraphQL/Definitions.hs
teh: changed.

PR #10, opened 2016-12-15
Mopup 2

teh:
Apologies, took more code than I expected to implement Union.

I removed your `Api.hs` because it blocked compiling and AFAICT we weren't going to use it?


src/GraphQL/Value.hs
```diff
@@ -79,24 +86,43 @@ instance ToJSON List where
...
+  toEncoding (Map xs) = pairs (fold (map (\(k, v) -> (toS k) .= v) xs))
 
 
 data Field = Field Name GraphQL.Value.Value deriving (Eq, Show, Ord)
 
 makeField :: (StringConv name Name, ToValue value) => name -> value -> Field
 makeField name value = Field (toS name) (toValue value)
 
+-- TODO I'm unclear about what we're going to use FieldSet for. Do we
+-- need it?
```
jml: I don't _think_ so. I think it was meant to be a realized selection set, but I can't be sure. Delete it, and if we discover we need it, we can add something similar & better later.

graphql-api.cabal
```diff
@@ -82,6 +81,9 @@ test-suite graphql-api-tests
     , tasty
     , tasty-hspec
   other-modules:
+      FileSystem
+      Git
+      T
```
jml: I'm not sure this change should be here.

src/GraphQL/TypeApi.hs
```diff
@@ -206,13 +219,68 @@ instance forall typeName interfaces fields m.
...
+
+  -- query "{ ... on Human { name } }"
+  -- [SelectionInlineFragment (InlineFragment (NamedType "Human") [] [SelectionField (Field "" "name" [] [] [])])]
+  buildResolver handler selection = do
+    -- GraphQL invariant is that all items in a Union must be objects
+    -- which means 1) they have fields 2) They are ValueMap
+    values <- map GValue.unionMap (traverse (runUnion @m @ru handler) selection)
+    case values of
+      Left _ -> implementationError "It looks like you used a non-Object type in a Union"
+      Right ok -> pure ok
```
jml: I've only skimmed over this whole diff hunk. I don't really understand it. I'll try to look at it later (post-merge) and come up with helpful comments / refactorings.

examples/UnionExample.hs
```diff
@@ -0,0 +1,32 @@
+module UnionExample where
```
teh: Hm, what's the best way to do that? It's not test code, and the only thing we care about is that it compiles.

src/GraphQL/Output.hs
```diff
@@ -1,3 +1,4 @@
+
```
teh: removed

src/GraphQL/TypeApi.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeInType #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE UndecidableInstances #-} -- for TypeError
```
teh: Without this we can't have nice TypeError messages. The worst case is that the type checker loops and dies when running out of stack. Compiling this code successfully ist the same as a unit test for this case. On top of that it's a closed type family so no one else can ever do anything to hit a type checker loop.

src/GraphQL/TypeApi.hs
```diff
@@ -142,17 +149,23 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
 -- Maybe we can use advanced fallbacks like these:
 -- https://wiki.haskell.org/GHC/AdvancedOverlap
 
+-- | Internal data type to capture a field's name + what to excute if
```
teh: done

examples/UnionExample.hs
jml: Put it in `tests` and make sure it's imported?

src/GraphQL/Value.hs
```diff
@@ -79,24 +86,43 @@ instance ToJSON List where
...
 -- return fields in query order as mandated by the spec.
-newtype Map = Map (Map.Map Name GraphQL.Value.Value) deriving (Eq, Ord, Show)
+newtype Map = Map [(Name,  GraphQL.Value.Value)] deriving (Eq, Ord, Show, Generic, Monoid)
+
+
+mapFromList :: [(Name,  GraphQL.Value.Value)] -> Map
+mapFromList = Map
+
+-- TODO this wouldbe nicer with a prism `_ValueMap` but don't want to
+-- pull in lens as dependency.
```
jml: > Lens dependency is somewhat contentious

Citation?

examples/UnionExample.hs
teh: but then it's not an examples directory any more :) - maybe a symlink?

jml: Can you demonstrate (maybe by creating a different branch & PR that runs on CI) that a failure to compile in examples will break the build on Circle?

teh: symlinked to tests instead, I think that's the easiest solution. I don't understand cabal enough to make it compile. PTAL

PR #17, opened 2016-12-15
Add support for doctests

### LINKED ISSUE #13: Add doctest to CI
Haskell doctest is pretty cool, because it means that examples in docs compile & execute as expected. We should add that, since we want to have examples.

### Links

* https://hackage.haskell.org/package/doctest
* https://hackage.haskell.org/package/doctest-discover
* https://github.com/sol/doctest
* https://ocharles.org.uk/blog/posts/2013-12-18-doctest.html
* https://www.reddit.com/r/haskell/comments/3zqv2e/using_doctestdiscover_with_stack/
* http://blog.rcook.org/blog/2016/doctest-discover-stack/
* http://blog.rcook.org/blog/2016/doctest-discover-stack-revisited/
* https://github.com/commercialhaskell/stack/issues/167

jml:
Doesn't actually add any doctests--that'll come later.

Review & merge #16 before reviewing this. Conflicts with #14.

Fixes #13.

### DIFF
--- a/graphql-api.cabal
+++ b/graphql-api.cabal
@@ -44,6 +44,28 @@ library
       GraphQL.Value
   default-language: Haskell2010
 
+test-suite graphql-api-doctests
+  type: exitcode-stdio-1.0
+  main-is: Doctests.hs
+  hs-source-dirs:
+      tests
+  default-extensions: NoImplicitPrelude OverloadedStrings RecordWildCards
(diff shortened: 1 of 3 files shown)


jml:
> Thanks! I'm not going to add doc-tests yet, will need to get a better feel for API and helper functions first.

I think they are a good way of acquiring that feel. Also a good way of structuring examples. But to each their own.

PR #24, opened 2016-12-16
Move Name to Value module and make invalid Objects impossible

jml:
I started out moving `Name` to the `Value` module with the intent of not adding any new TODOs. I ended up adding three, with a much bigger PR than I expected.

That's because I tried to eliminate the TODO that asked `What the heck does unionObject do?` I wrote a test to figure it out and found that it constructed invalid objects--objects with duplicate field names. I really don't want to ever have that bug again, so I jotted down a property and updated the types to give `Object` smart constructors so it's now only possible to construct Objects with unique field names.

To be honest, I still don't understand why `unionObjects` is the right behaviour in the union resolver, but I'm certain that its old behaviour was definitely wrong.

* Start of tests for `GraphQL.Value`
* Remove duplication of `Name` from `Schema` and `Value`, moving everything to `Value`
* Create `ObjectField` type for object fields
* Rename `Map` to `Object`, as per the TODO comment and spec
* Change some `data` to `newtype` (hlint warning, removes runtime overhead)
* Remove unnecessary parens (hlint)
* Extract `wrongType` helper (we can have top-level private functions now that we have explicit export lists)

### DIFF
--- a/src/GraphQL/TypeApi.hs
+++ b/src/GraphQL/TypeApi.hs
@@ -210,7 +210,7 @@ class RunFields m a where
   type RunFieldsType m a :: Type
   -- Runfield is run on a single QueryTerm so it can only ever return
   -- one (Text, Value)
-  runFields :: RunFieldsType m a -> AST.Selection -> m (Text, GValue.Value)
+  runFields :: RunFieldsType m a -> AST.Selection -> m GValue.ObjectField
 
 
 instance forall f fs m.
@@ -223,14 +223,15 @@ instance forall f fs m.
   runFields (lh :<> rh) selection@(AST.SelectionField (AST.Field alias name _ _ _)) =
     let NamedFieldExecutor k mValue = buildFieldResolver @m @f lh selection
     in case name == k of
+      False -> runFields @m @fs rh selection
       True -> do
         -- execute action to retrieve field value
         value <- mValue
         -- NB "alias" is encoded in-band. It cannot be set to empty in
         -- a query so the empty value means "no alias" and we use the
         -- name instead.
-        pure (if alias == "" then name else alias, value)
-      False -> runFields @m @fs rh selection
+        let name' = GValue.Name $ if alias == "" then name else alias
+        pure (GValue.ObjectField name' value)
 
   runFields _ f = queryError ("Unexpected Selection value. Is the query normalized?: " <> show f)
 
@@ -252,7 +253,7 @@ instance forall typeName interfaces fields m.
     -- We're evaluating an Object so we're collecting (name, Value)
     -- pairs from runFields and build a GValue.Map with them.
     r <- forM selectionSet $ \selection -> runFields @m @fields handler selection
-    pure $ GValue.toValue (GValue.objectFromList r)
+    pure $ GValue.toValue (GValue.Object r)
 
 
 -- | Closed type family to enforce the invariant that Union types
--- a/src/GraphQL/TypeApi.hs
+++ b/src/GraphQL/TypeApi.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE AllowAmbiguousTypes #-}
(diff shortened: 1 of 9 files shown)


src/GraphQL/TypeApi.hs
```diff
@@ -245,40 +250,47 @@ instance forall typeName interfaces fields m.
   buildResolver mHandler selectionSet = do
     -- First we run the actual handler function itself in IO.
     handler <- mHandler
-    -- We're evaluating an Object so we're collecting (name, Value)
-    -- pairs from runFields and build a GValue.Map with them.
-    r <- forM selectionSet $ \selection -> runFields @m @fields handler selection
-    pure $ GValue.toValue (GValue.mapFromList r)
+    -- We're evaluating an Object so we're collecting ObjectFields from
```
teh: FWIW I think we should catch duplicate-field errors on query validation, not query evaluation. But we can keep the code for now. Maybe add a TODO

src/GraphQL/TypeApi.hs
```diff
@@ -245,40 +250,47 @@ instance forall typeName interfaces fields m.
...
          , MonadThrow m
          , RunFields m fields
          , KnownSymbol typeName
-         ) => RunUnion m ((Object typeName interfaces fields):rest) where
+         ) => RunUnion m (Object typeName interfaces fields:rest) where
   runUnion (lh :<|> rh) fragment@(AST.SelectionInlineFragment (AST.InlineFragment (AST.NamedType queryTypeName) [] subSelection))
-    | typeName == queryTypeName = buildResolver @m @(Object typeName interfaces fields) lh subSelection
+    | typeName == queryTypeName = do
+        result <- buildResolver @m @(Object typeName interfaces fields) lh subSelection
+        -- TODO: See if we can prevent this from happening at compile time.
```
teh: I think we can using something like: https://hackage.haskell.org/package/type-level-sets-0.8.0.0/docs/Data-Type-Set.html . Does increase compile times substantially though and the errors are not very readable last time I tried. (unactionable)

src/GraphQL/Value.hs
```diff
@@ -44,53 +63,54 @@ instance ToJSON GraphQL.Value.Value where
...
-    isValueMap = \case
-      (ValueMap m) -> Right m
-      _ -> Left "non-ValueMap member"
+    fieldNames = [name | ObjectField name _ <- fields]
+
+objectFromList :: [(Name, Value)] -> Maybe Object
+objectFromList = makeObject . map (uncurry ObjectField)
 
+unionObjects :: [Object] -> Maybe Object
+unionObjects = makeObject . (=<<) objectFields
```
teh: I'm not sure what's going on here, is it possible to expand this line bit?

src/GraphQL/TypeApi.hs
```diff
@@ -245,40 +250,47 @@ instance forall typeName interfaces fields m.
...
+        case GValue.toObject result of
+          Nothing -> queryError $ "Expected object as result of union query: " <> show result
```
teh: this is not a queryError, it's an implementation error in our library so should panic IMO.

src/GraphQL/TypeApi.hs
```diff
@@ -130,32 +131,35 @@ lookupValue name args = case find (\(AST.Argument name' _) -> name' == name) arg
...
 
 instance forall v. ReadValue v => ReadValue [v] where
   readValue (AST.ValueList (AST.ListValue values)) = traverse (readValue @v) values
-  readValue v = Left ("Not a List:" <> (show v))
+  readValue v = wrongType "List" v
 
 instance forall v. ReadValue v => ReadValue (Maybe v) where
   valueMissing _ = pure Nothing
-  readValue v = map Just (readValue @v v)
+  readValue v = map Just (readValue v)
```
teh: I know these type applications aren't strictly necessary but they do often lead to better compiler errors because they help ghc constrain things early. That's why I stick them in

jml: Restored. (Although I'm pretty sure ghc knows that the type of `v` is `v` in this case, so not sure how it could give better errors to say so).

src/GraphQL/TypeApi.hs
jml: The code that merges objects still needs to do _something_ when it gets duplicate fields. Am happy to add a TODO but don't know what that TODO would be.

src/GraphQL/TypeApi.hs
jml: Changed

src/GraphQL/Value.hs
jml: Changed to pointful style:

```haskell
unionObjects objects = makeObject (objects >>= objectFields)
```

That clearer? Can use `concatMap` instead of `>>=` if you'd prefer:

```haskell
unionObjects objects = makeObject (concatMap objectFields objects)
```

src/GraphQL/TypeApi.hs
jml: Well, I was thinking more that we could have a separate resolver typeclass for things that are guaranteed to resolve to an object. Then the current resolver typeclass would be a superclass of that (if we know how to interpret a thing as an object, we know how to interpret it as a value). 

Type-level sets would presumably be for enforcing uniqueness.

jml:
PTAL

src/GraphQL/TypeApi.hs
teh: Sure, I meant that if we validate the query there would be no need to check for duplicate fields. Even if the Object defines duplicate fields (which is another interesting bug we should catch, see discussion below) then we'd never hit this condition.

jml:
> Sure, I meant that if we validate the query there would be no need to check for duplicate fields. Even if the Object defines duplicate fields (which is another interesting bug we should catch, see discussion below) then we'd never hit this condition.

Ah, yes. I make a comment about this in #26.

PR #25, opened 2016-12-17
Drop input values

jml:


src/GraphQL/Schema.hs
```diff
@@ -119,26 +121,11 @@ newtype TypeExtensionDefinition = TypeExtensionDefinition ObjectTypeDefinition
...
-  | ValueBoolean Bool
-  | ValueString Text
-  | ValueEnum Name
-  | ValueList ListValue
-  | ValueObject [ObjectField] deriving (Eq, Show)
-
-data ListValue = ListValue [InputValue] deriving (Eq, Show)
-data ObjectValue = ObjectValue [ObjectField] deriving (Eq, Show)
-data ObjectField = ObjectField Name InputValue deriving (Eq, Show)
+type DefaultValue = Value
```
teh: What's the use case for `DefaultValue`? Can you add a comment (doesn't need to be a full doc comment, just a note what it's purpose will be)

Issue #23, opened 2016-12-16
QuickCheck tests for value correctness


jml:
All the stuff in `GraphQL.Value` should have generators and quickcheck tests. It's pretty fundamental, but a lot of work.

@teh Do you have any opinion about whether we should put `Arbitrary` instances in the library? My current feeling is custom generators in the tests, e.g.

```haskell
validName :: Gen Name
validName = Name <$> arbitrary

validObject :: Gen Object
-- and so forth ...
```

PR #26, opened 2016-12-17
Smart constructor for Name

jml:
The quest to remove TODOs never ends. Although this removes one, it makes another.

Builds on https://github.com/jml/graphql-api/pull/24, so review & merge that before reviewing this (or just review https://github.com/jml/graphql-api/pull/26/commits/ddbd8afbb4723f15236a039618a09a925cec0946).

Was going to do QuickCheck instances in this as per #23 but this turned out to be more complicated than expected and also now I really have to go.

## Smart constructor for Name

Name is restricted in value, so when we have something that accepts a `Name` as a parameter, it should be able to trust that it's valid.

Stopped exporting `Name` constructor and now export two constructors:

 1. `makeName`
 2. `unsafeMakeName`

They do what they say on the tin.

There are three use-cases for `unsafeMakeName`:

 1. In tests, where we are passing a literal.
 2. In `Definitions`, because we cannot get GHC to enforce a restricted set of values for a `Symbol`.
 3. In `TypeApi`, because we're operating on Data.GraphQL.AST directly when we should be operating on a pre-restricted type. Noted as a TODO comment.

I tried adding a `HasName` type class to `Definitions`, but couldn't figure it out in the ten minutes I gave myself.

## Misc cleanups

* Dropped superfluous parens

### DIFF
--- a/src/GraphQL/TypeApi.hs
+++ b/src/GraphQL/TypeApi.hs
@@ -230,7 +230,14 @@ instance forall f fs m.
         -- NB "alias" is encoded in-band. It cannot be set to empty in
         -- a query so the empty value means "no alias" and we use the
         -- name instead.
-        let name' = GValue.Name $ if alias == "" then name else alias
+
+        -- TODO: We need to use 'unsafeMakeName' here (which might panic)
+        -- because our API is currently written in terms of the Data.GraphQL
+        -- parser, which provides no type-level guarantees of name safety. We
+        -- should instead have our APIs in terms of 'Canonicalquery' (and the
+        -- rest of 'Data.GraphQL.Input', not yet written), and have that be
+        -- responsible for rejecting queries with invalid names.
+        let name' = GValue.unsafeMakeName $ if alias == "" then name else alias
         pure (GValue.ObjectField name' value)
 
   runFields _ f = queryError ("Unexpected Selection value. Is the query normalized?: " <> show f)
--- a/src/GraphQL/Value.hs
+++ b/src/GraphQL/Value.hs
@@ -7,7 +7,9 @@ module GraphQL.Value
   ( Value(..)
   , toObject
   , ToValue(..)
-  , Name(Name)
+  , Name
+  , makeName
+  , unsafeMakeName
   , List
   , String
     -- * Objects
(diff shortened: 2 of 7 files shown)


teh:
Looking at commits looks fine but would prefer to see rebase on other branch if you have time.

jml:
It has been rebased?

jml:
Or I guess approve #24 and I can rebate on master after merging.

jml:
Sorry about rebase. Had a brain fart. Done now.

Also removed the `IsString` instance from `Name`, as it was allowing constructing invalid names. We could re-introduce it and have it panic, which wouldn't be so bad as it would only ever be called on literals. I thought about that and thought that probably it would be less surprising for users if they had to be explicit about it.

src/GraphQL/TypeApi.hs
```diff
@@ -230,7 +230,14 @@ instance forall f fs m.
...
         -- name instead.
-        let name' = GValue.Name $ if alias == "" then name else alias
+
+        -- TODO: We need to use 'unsafeMakeName' here (which might panic)
+        -- because our API is currently written in terms of the Data.GraphQL
+        -- parser, which provides no type-level guarantees of name safety. We
+        -- should instead have our APIs in terms of 'Canonicalquery' (and the
+        -- rest of 'Data.GraphQL.Input', not yet written), and have that be
+        -- responsible for rejecting queries with invalid names.
+        let name' = GValue.unsafeMakeName $ if alias == "" then name else alias
```
teh: style feedback: for subexpressions I find `( .. )` more readable than `$`, but no need to change.

src/GraphQL/Value.hs
```diff
@@ -7,36 +7,62 @@ module GraphQL.Value
...
+makeName = map Name . hush . parseOnly Parser.name
+
+-- | Create a 'Name', panicking if the given text is invalid.
+--
+-- Prefer 'makeName' to this in all cases.
+--
+-- >>> makeName "foo"
+-- Just (Name {getName = "foo"})
+-- >>> makeName "9-bar"
+-- Nothing
```
teh: I think this panics instead of returning Nothing?

jml: Good catch! Forgot to rename when adjusting comment, so doctests didn't catch this. Fixed.

src/GraphQL/TypeApi.hs
jml: I find the opposite, I'm afraid. Will try to blend in though.

PR #27, opened 2016-12-17
QuickCheck for Values

### LINKED ISSUE #23: QuickCheck tests for value correctness

jml:
Pretty boring (at last). Depends on #26.

Fixes #23.

Note that `List` is too generously typed. I can't think of a good way of encoding homogeneous lists without adding a type parameter. I think we'll have to suck that one up.

### DIFF
--- a/graphql-api.cabal
+++ b/graphql-api.cabal
@@ -33,6 +33,7 @@ library
     , aeson
     , containers
     , graphql
+    , QuickCheck
     , text
   exposed-modules:
       GraphQL.Definitions
@@ -84,6 +85,7 @@ test-suite graphql-api-tests
     , containers
     , graphql
     , graphql-api
+    , hspec
     , tasty
     , tasty-hspec
   other-modules:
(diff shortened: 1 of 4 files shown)


jml:
Reckon I should do this for all of the arbitrary instances? I could go
either way between Arbitrary and named generators.

teh:
No what you have is good for now, we can add the extra `forAll` code as we need it.

jml:
Thanks. 

I thought of a way we might be able to restructure the type, at least to make sure that the "top-level" types are the same (e.g. everything is an "Object" but cannot enforce that they are the same type of object). I think that's the best we can get since this level is entirely separate from the GraphQL type system, and that's as it should be.

I'll file a ticket though, and move on with `CanonicalQuery`.

PR #31, opened 2016-12-19
No more ()

teh:
This is an attempt to remove the `:<> ()` part in our handlers.

### DIFF
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -7,7 +7,7 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeInType #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE MultiParamTypeClasses, TypeFamilyDependencies #-}
 {-# LANGUAGE UndecidableInstances #-} -- for TypeError
 
 module GraphQL.Server
@@ -182,12 +182,14 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
 -- name doesn't match.
 data NamedFieldExecutor m = NamedFieldExecutor AST.Name (m GValue.Value)
 
+type family FieldHandler m (a :: Type) :: Type where
+  FieldHandler m (Field ks t) = Handler m t
+  FieldHandler m (Argument ks t :> f) = t -> FieldHandler m f
+
 class (MonadThrow m, MonadIO m) => BuildFieldResolver m a where
-  type FieldHandler m a :: Type
   buildFieldResolver :: FieldHandler m a -> AST.Selection -> NamedFieldExecutor m
 
 instance forall ks t m. (KnownSymbol ks, HasGraph m t, MonadThrow m, MonadIO m) => BuildFieldResolver m (Field ks t) where
-  type FieldHandler m (Field ks t) = Handler m t
   buildFieldResolver handler (AST.SelectionField (AST.Field _ _ _ _ selectionSet)) =
     let childResolver = buildResolver @m @t handler selectionSet
         name = toS (symbolVal (Proxy :: Proxy ks))
@@ -197,7 +199,6 @@ instance forall ks t m. (KnownSymbol ks, HasGraph m t, MonadThrow m, MonadIO m) 
 
 
 instance forall ks t f m. (MonadThrow m, KnownSymbol ks, BuildFieldResolver m f, ReadValue t) => BuildFieldResolver m (Argument ks t :> f) where
-  type FieldHandler m (Argument ks t :> f) = t -> FieldHandler m f
   buildFieldResolver handler selection@(AST.SelectionField (AST.Field _ _ arguments _ _)) =
     let argName = toS (symbolVal (Proxy :: Proxy ks))
         v = maybe (valueMissing @t argName) (readValue @t) (lookupValue argName arguments)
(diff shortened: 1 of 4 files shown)


jml:
Can you rebase this & get tests passing before I look?

src/GraphQL/Server.hs
```diff
@@ -248,24 +258,52 @@ instance forall f fs m.
...
+instance forall ks t m.
+         ( BuildFieldResolver m (Field ks t)
+         , MonadThrow m
+         , MonadIO m
+         ) => RunFields m (Field ks t) where
+  type RunFieldsHandler m (Field ks t) = FieldHandler m (Field ks t)
+  runFields lh selection@(AST.SelectionField (AST.Field alias name _ _ _)) =
+    let NamedFieldExecutor k mValue = buildFieldResolver @m @(Field ks t) lh selection
+    in case name == k of
+      False -> queryError ("Query for undefined selection:" <> show selection)
```
teh: I'm unhappy about the duplication in `RunFields` instances but I think we can reduce it a bit with helper functions.

src/GraphQL/Server.hs
```diff
@@ -182,12 +182,14 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
 -- name doesn't match.
 data NamedFieldExecutor m = NamedFieldExecutor AST.Name (m GValue.Value)
 
+type family FieldHandler m (a :: Type) :: Type where
```
jml: Can you please add a doc comment?

src/GraphQL/Server.hs
```diff
@@ -208,22 +209,31 @@ instance forall ks t f m. (MonadThrow m, KnownSymbol ks, BuildFieldResolver m f,
...
+  RunFieldsType m ((Field ks t) ': rest) = Field ks t :<> RunFieldsType m rest
+  RunFieldsType m ((a :> b) ': rest) = (a :> b) :<> RunFieldsType m rest
+  RunFieldsType m a = TypeLits.TypeError (
+    'TypeLits.Text "All field entries in an Object must be Field or Argument :> Field. Got: " 'TypeLits.:<>: 'TypeLits.ShowType a)
+
 
 class RunFields m a where
-  type RunFieldsType m a :: Type
+  type RunFieldsHandler m a :: Type
   -- Runfield is run on a single QueryTerm so it can only ever return
```
jml: Please get case of `Runfield` correct (I think it's supposed to be `runFields` but am not sure).

src/GraphQL/Server.hs
```diff
@@ -208,22 +209,31 @@ instance forall ks t f m. (MonadThrow m, KnownSymbol ks, BuildFieldResolver m f,
...
 instance forall f fs m.
          ( BuildFieldResolver m f
          , RunFields m fs
-         ) => RunFields m (f:fs) where
-  type RunFieldsType m (f:fs) = (FieldHandler m f) :<> RunFieldsType m fs
+         , MonadThrow m
+         , MonadIO m
+         ) => RunFields m (f :<> fs) where
+  type RunFieldsHandler m (f :<> fs) = FieldHandler m f :<> RunFieldsHandler m fs
+--  type RunFieldsType m (f:fs) = (FieldHandler m f) :<> RunFieldsType m fs
```
jml: Remove commented out code please.

src/GraphQL/Server.hs
```diff
@@ -274,22 +312,27 @@ instance forall typeName interfaces fields m.
 -- | Closed type family to enforce the invariant that Union types
 -- contain only Objects.
 type family RunUnionType m (a :: [Type]) :: Type where
-  RunUnionType m (Object typeName interfaces fields:rest) = Handler m (Object typeName interfaces fields) :<|> RunUnionType m rest
-  RunUnionType m '[] = ()
+  RunUnionType m '[Object typeName interfaces fields] = Object typeName interfaces fields
+  RunUnionType m (Object typeName interfaces fields:rest) = Object typeName interfaces fields :<|> RunUnionType m rest
```
jml: Surely that should be `field` singular?

tests/Examples/UnionExample.hs
```diff
@@ -17,13 +17,13 @@ type O2 = Object "O2" '[] '[Field "o2" Text]
...
-o1 = pure ((pure "hello from O1") :<> ())
+o1 = pure (pure "hello from O1")
 
 o2 :: Handler IO O2
-o2 = pure ((pure "hello from O2") :<> ())
+o2 = pure (pure "hello from O2")
 
 tHandler :: Handler IO T
-tHandler = o1 :<|> o2 :<|> ()
+tHandler = o1 :<|> o2
```
jml: Can you please add an example where you demonstrate that handlers are monoidal?

e.g. for `o1`, `o2`, `o3`, that:

```haskell
tHandler = (o1 :<|> o2) :<|> e3
```

and 

```haskell
tHandler = o1 :<|> (o2 :<|> o3)
```

Both compile?

No need to show they are equal.

src/GraphQL/Server.hs
teh: Done

src/GraphQL/Server.hs
teh: Fixed, also fixed rest of comment

src/GraphQL/Server.hs
teh: done

src/GraphQL/Server.hs
teh: I looked around but not sure where - do you have a link?

src/GraphQL/Server.hs
teh: Ah, this is actually `((Object typeName interfaces fields):rest)` - I can add the parens but I saw you started removing my extra parens that I left in so I removed them as well. Up to you which style we follow I don't mind.

tests/Examples/UnionExample.hs
teh: These operators (`:<|>` and `:<>`) are **not** associative! That's the reason we can't have the `fieldList` function. They construct a structure that's isomorphic to a right-nested tuple. Any re-nesting of that tuple will result in a different type.

It might possible to lift that restriction at some point but it's beyond my capabilities ATM.

src/GraphQL/Server.hs
jml: Let's stick without parens for now. That's what `hlint` does, and it's a de facto standard.

tests/Examples/UnionExample.hs
jml: Hmm! Thanks.

teh:
Fixed tutorial docs as well, would indeed be cool to have them in CI so I don't have to remember. servant uses tinc to manage dependencies, but maybe we can use stack?

PR #37, opened 2016-12-21
Roundtrip strings

jml:
Turns out graphql doesn't support escaped strings. This makes it do so by re-using Aeson's code, on the assumption that GraphQL strings are identical to JSON strings.


jml:
> LGTM, had a slight rewrite of string parser but doesn't matter this works fine (must resist fiddling!)

Will merge now, but curious as to the gist of this.

PR #38, opened 2016-12-21
WIP: Full value roundtrip

jml:
#37 does this for strings, this does it for all values. Review & merge #37 first to see a good diff for this.

Adds quickcheck tests for roundtripping values. They currently fail, but not on every run. I've added a test that exercises the broken path, but I haven't had time to minimize.

It might be useful to begin by adding better shrinking support to the `Arbitrary` instances we already have. That's what I'm going to do next time I face quickcheck failures.

### DIFF
--- a/graphql-api.cabal
+++ b/graphql-api.cabal
@@ -89,6 +89,7 @@ test-suite graphql-api-tests
     , containers
     , graphql-api
     , hspec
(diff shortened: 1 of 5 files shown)


jml:
LGTM. I think you have to approve.

PR #40, opened 2016-12-22
Mixed bag

jml:
Sorry, this is a bit of a mixed bag of changes that I've been working on in my spare time over the last few weeks. I figure I'd better get them in now, even though none of them are exactly whole changes.

* quickcheck tests for `findDuplicates`, because it's always worth checking
* benchmarks for `findDuplicates`. we don't care about it quite so much, but it's nice to start with _something_
* move all the name parsing machinery to the `AST` module in preparation for changing the `Name` type
* add a bunch of tests to show how the parser behaves (this is my equivalent of @teh pasting stuff into a repl)
* change `buildFieldResolver` to return an explicit error via `Either` if it somehow matches on the wrong pattern
* extract common code from the `RunFields` type class. At this point, the implementations only differ in how they fallback if there's no match and what type variables they use. I speculate that we might be able to tighten up the typeclass definition so they just provide those.
* tidy of the TypeApiTests so they duplicate less code and provide a slightly more functional working example of using the API. (I guess we should rename this module).

### DIFF
--- a/benchmarks/Main.hs
+++ b/benchmarks/Main.hs
@@ -1,0 +1,12 @@
+module Main (main) where
+
+import Protolude
+
+import Criterion.Main (bgroup, defaultMain)
+import qualified Validation
+
+
+main :: IO ()
+main = do
+  defaultMain [ bgroup "GraphQL API" Validation.benchmarks
+              ]
--- a/src/GraphQL/Internal/AST.hs
+++ b/src/GraphQL/Internal/AST.hs
@@ -1,5 +1,6 @@
 module GraphQL.Internal.AST
   ( Name
+  , pName
   , Document(..)
   , Definition(..)
   , OperationDefinition(..)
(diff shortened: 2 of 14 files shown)


benchmarks/Main.hs
```diff
@@ -0,0 +1,12 @@
+module Main (main) where
+
+import Protolude
+
+import Criterion.Main (bgroup, defaultMain)
+import qualified Validation
+
+
```
teh: +1

src/GraphQL/Internal/AST.hs
```diff
@@ -1,9 +1,11 @@
 module GraphQL.Internal.AST
   ( Name
+  , pName
```
teh: would you mind renaming to `parseName` or similar? Took me a while to figure out what you mean here.

src/GraphQL/Server.hs
```diff
@@ -181,33 +181,54 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
...
+executeNamedField (AST.Field alias name _ _ _) (NamedFieldExecutor k mValue)
+  | name == k = do
+      value <- mValue
+      let name' = GValue.unsafeMakeName $ if alias == "" then name else alias
+      pure (Just (GValue.ObjectField name' value))
+  | otherwise = pure Nothing
+
+-- Deconstruct object type signature and handler value at the same
+-- time and run type-directed code for each field.
+resolveField :: forall a (m :: * -> *). BuildFieldResolver m a => m GValue.ObjectField -> FieldHandler m a -> AST.Selection -> m GValue.ObjectField
```
teh: our convention is to use `Type` instead of `*`

src/GraphQL/Internal/AST.hs
jml: Renamed to `nameParser`, as I would expect a function `parseName` to return a `Name` not a parser.

src/GraphQL/Server.hs
jml: Ahhhh, so _that's_ what `Type` means. I'm familiar with `*` from the repl & compilation errors, and had assumed that `Type` was something special related to all the magic you're doing. :smiling_imp: 

Changed to use `Type`.

FWIW, this signature was added with a keystroke, using the 'apply suggestion' feature of intero, which looks at the compiler warning/error and applies the suggestion.

PR #41, opened 2016-12-22
Change error handling behaviour

jml:
I got this down to a fairly minimal change. I don't have time right now to figure out why this changes both error messages, but I suspect something to do with MonadThrow.


jml:
I think I figured out what's going on. I could fix it, but I'll wait for your error handling change to land first.

PR #43, opened 2016-12-30
Move error handling closer to the spec

teh:
The main change here is that instead of an `Either` we now have a `Result` which accumulates errors and values at the same time.

* This is still a draft but wanted to get your opinion early 
* I'm not super happy with `ResolveFieldResult` but not sure ATM how to phrase better (other than maybe a separate `data` type)
* MonadIO & throw are gone
* Error accumulation needs a nicer representation but a I can't tell from the [description](https://facebook.github.io/graphql/#sec-Errors) what errors for specific fields should look like.

### DIFF
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -88,7 +88,7 @@ ok v = Result [] v
 
 -- TODO instead of SelectionSet we want something like
 -- NormalizedSelectionSet which has query fragments etc. resolved.
-class (MonadIO m) => HasGraph m a where
+class (Monad m) => HasGraph m a where
   type Handler m a
   buildResolver :: Handler m a -> CanonicalQuery -> m Result
 
@@ -110,24 +110,24 @@ class ReadValue a where
 -- TODO not super hot on individual values having to be instances of
 -- HasGraph but not sure how else we can nest either types or
 -- (Object _ _ fields). Maybe instead of field we need a "SubObject"?
-instance forall m. (MonadIO m) => HasGraph m Int32 where
+instance forall m. (Monad m) => HasGraph m Int32 where
   type Handler m Int32 = m Int32
   -- TODO check that selectionset is empty (we expect a terminal node)
   buildResolver handler _ =  do
     map (ok . GValue.toValue) handler
 
-instance forall m. (MonadIO m) => HasGraph m Double where
+instance forall m. (Monad m) => HasGraph m Double where
   type Handler m Double = m Double
   -- TODO check that selectionset is empty (we expect a terminal node)
   buildResolver handler _ =  map (ok . GValue.toValue) handler
 
-instance forall m. (MonadIO m) => HasGraph m Text where
+instance forall m. (Monad m) => HasGraph m Text where
   type Handler m Text = m Text
   -- TODO check that selectionset is empty (we expect a terminal node)
   buildResolver handler _ =  map (ok . GValue.toValue) handler
 
 
-instance forall m hg. (MonadIO m, HasGraph m hg) => HasGraph m (List hg) where
+instance forall m hg. (Monad m, HasGraph m hg) => HasGraph m (List hg) where
   type Handler m (List hg) = [Handler m hg]
   buildResolver handler selectionSet = map resultValues a
     where
@@ -140,7 +140,7 @@ instance forall m hg. (MonadIO m, HasGraph m hg) => HasGraph m (List hg) where
         let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
         in Result errs (GValue.toValue valueList)
 
-instance forall m ks enum. (MonadIO m, GraphQLEnum enum) => HasGraph m (Enum ks enum) where
+instance forall m ks enum. (Monad m, GraphQLEnum enum) => HasGraph m (Enum ks enum) where
   type Handler m (Enum ks enum) = enum
   buildResolver handler _ = (pure . ok . enumToValue) handler
 
@@ -235,10 +235,10 @@ type family FieldHandler m (a :: Type) :: Type where
   FieldHandler m (Field ks t) = Handler m t
   FieldHandler m (Argument ks t :> f) = t -> FieldHandler m f
 
-class (MonadIO m) => BuildFieldResolver m a where
+class (Monad m) => BuildFieldResolver m a where
   buildFieldResolver :: FieldHandler m a -> AST.Selection -> Either QueryError (NamedValueResolver m)
 
-instance forall ks t m. (KnownSymbol ks, HasGraph m t, HasAnnotatedType t, MonadIO m) => BuildFieldResolver m (Field ks t) where
+instance forall ks t m. (KnownSymbol ks, HasGraph m t, HasAnnotatedType t, Monad m) => BuildFieldResolver m (Field ks t) where
   buildFieldResolver handler (AST.SelectionField (AST.Field _ _ _ _ selectionSet)) = do
     let resolver = buildResolver @m @t handler selectionSet
     field <- first (QueryError. AST.formatNameError) (getFieldDefinition @(Field ks t))
@@ -289,7 +289,7 @@ class RunFields m a where
 instance forall f fs m.
          ( BuildFieldResolver m f
          , RunFields m fs
-         , MonadIO m
+         , Monad m
          ) => RunFields m (f :<> fs) where
   runFields (handler :<> nextHandlers) selection =
     resolveField @f @m handler nextHandler selection
@@ -298,7 +298,7 @@ instance forall f fs m.
 
 instance forall ks t m.
          ( BuildFieldResolver m (Field ks t)
-         , MonadIO m
+         , Monad m
          ) => RunFields m (Field ks t) where
   runFields handler selection =
     resolveField @(Field ks t) @m handler nextHandler selection
@@ -307,7 +307,7 @@ instance forall ks t m.
 
 instance forall m a b.
          ( BuildFieldResolver m (a :> b)
-         , MonadIO m
+         , Monad m
          ) => RunFields m (a :> b) where
   runFields handler selection =
     resolveField @(a :> b) @m handler nextHandler selection
@@ -317,7 +317,7 @@ instance forall m a b.
 
 instance forall typeName interfaces fields m.
          ( RunFields m (RunFieldsType m fields)
-         , MonadIO m
+         , Monad m
          ) => HasGraph m (Object typeName interfaces fields) where
   type Handler m (Object typeName interfaces fields) = m (RunFieldsHandler m (RunFieldsType m fields))
 
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -121,6 +121,10 @@ instance Defaultable Bool
 
 instance Defaultable Text
 
+instance Defaultable (Maybe a) where
+  -- | The default for @Maybe a@ is @Nothing@.
+  defaultFor _ = pure Nothing
+
 -- TODO not super hot on individual values having to be instances of
 -- HasGraph but not sure how else we can nest either types or
 -- (Object _ _ fields). Maybe instead of field we need a "SubObject"?
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -203,7 +203,7 @@ data NamedValueResolver m = NamedValueResolver AST.Name (m Result)
 -- Iterate through handlers (zipped together with their type
 -- definition) and execute handler if the name matches.
 --
--- TODO: tuple return not great (but either is wrong, too because it discards errors. Also probably need Maybe for ObjectField?
+-- TODO: tuple return not great (but either is wrong, too because it discards errors.
 type ResolveFieldResult = ([QueryError], Maybe GValue.ObjectField)
 resolveField :: forall a (m :: Type -> Type). BuildFieldResolver m a
   => FieldHandler m a -> m ResolveFieldResult -> AST.Selection -> m ResolveFieldResult
@@ -244,8 +244,8 @@ instance forall ks t m. (KnownSymbol ks, HasGraph m t, HasAnnotatedType t, Monad
     field <- first (QueryError. AST.formatNameError) (getFieldDefinition @(Field ks t))
     let name = getName field
     Right (NamedValueResolver name resolver)
-  buildFieldResolver _ _ =
-    panic "this can not happen"
+  buildFieldResolver _ f =
+    Left (InternalError ("Unexpected query selection in buildFieldResolver: " <> show f))
 
 
 instance forall ks t f m.
@@ -259,8 +259,8 @@ instance forall ks t f m.
     let argName = getName argument
     value <- first QueryError (maybe (valueMissing @t argName) (readValue @t) (lookupValue argName arguments))
     buildFieldResolver @m @f (handler value) selection
-  buildFieldResolver _ _ =
-    panic "this can not happen"
+  buildFieldResolver _ f =
+    Left (InternalError ("Unexpected query selection in buildFieldResolver: " <> show f))
 
 
 -- We only allow Field and Argument :> Field combinations:
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -77,8 +77,15 @@ infixr 8 :<|>
 -- Result collects errors and values at the same time unless a handler
 -- tells us to bail out in which case we stop the processing
 -- immediately.
--- TODO: Make QueryError a Monoid and aggregate according to that.
 data Result a = Result [ResolverError] a deriving (Show, Functor)
+
+-- Aggregating results keeps all errors and creates a ValueList
+-- containing the individual values.
+aggregateResults :: [Result GValue.Value] -> Result GValue.Value
+aggregateResults r =
+  let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
+  in Result errs (GValue.toValue valueList)
+
 
 instance Applicative Result where
   pure v = Result [] v
@@ -152,16 +159,9 @@ instance forall m. (Monad m) => HasGraph m Text where
 
 instance forall m hg. (Monad m, HasGraph m hg) => HasGraph m (List hg) where
   type Handler m (List hg) = [Handler m hg]
-  buildResolver handler selectionSet = map resultValues a
+  buildResolver handler selectionSet = map aggregateResults a
     where
       a = traverse (flip (buildResolver @m @hg) selectionSet) handler
-      -- Aggregate results. TODO: Result should probably be a monoid
-      -- so we can collect values but I'm not sure what the monoid of
-      -- Value itself is.
-      resultValues :: [Result GValue.Value] -> Result GValue.Value
-      resultValues r =
-        let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
-        in Result errs (GValue.toValue valueList)
 
 instance forall m ks enum. (Monad m, GraphQLEnum enum) => HasGraph m (Enum ks enum) where
   type Handler m (Enum ks enum) = enum
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -82,9 +82,7 @@ data Result a = Result [ResolverError] a deriving (Show, Functor, Eq)
 -- Aggregating results keeps all errors and creates a ValueList
 -- containing the individual values.
 aggregateResults :: [Result GValue.Value] -> Result GValue.Value
-aggregateResults r =
-  let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
-  in Result errs (GValue.toValue valueList)
+aggregateResults r = GValue.toValue <$> sequenceA r
 
 instance Applicative Result where
   pure v = Result [] v
@@ -315,8 +313,8 @@ instance forall typeName interfaces fields m.
     -- We're evaluating an Object so we're collecting ObjectFields from
     -- runFields and build a GValue.Map with them.
     r <- forM selectionSet (runFields @m @(RunFieldsType m fields) handler)
-    let (errs, fields) = foldr' (\(Result ea fa) (eb, fbs) -> (eb <> ea, fa:fbs)) ([], []) r
-
-    case GValue.makeObject (catMaybes fields) of
+    -- let (errs, fields) = foldr' (\(Result ea fa) (eb, fbs) -> (eb <> ea, fa:fbs)) ([], []) r
+    let (Result errs obj)  = GValue.makeObject . catMaybes <$> sequenceA r
+    case obj of
       Nothing -> pure (Result [InvalidQueryError ("Duplicate fields in set: " <> show r)] GValue.ValueNull)
       Just object -> pure (Result errs (GValue.ValueObject object))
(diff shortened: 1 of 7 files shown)


tests/TypeApiTests.hs
```diff
@@ -43,26 +43,26 @@ getQuery query =
         parseOnly (document <* endOfInput) query
   in selectionSet
 
-runQuery :: AST.SelectionSet -> IO (Either Text Value)
+runQuery :: AST.SelectionSet -> IO (Either Text Result)
```
jml: What would `Left` represent here?

tests/TypeApiTests.hs
```diff
@@ -17,9 +17,9 @@ import GraphQL.Server
   , QueryError(..)
   , buildResolver
   , (:<>)(..)
+  , Result(..)
```
jml: Where's this defined? I don't see it in this diff.

jml:
> Error accumulation needs a nicer representation but a I can't tell from the description what errors for specific fields should look like.

This part is already encoded in GraphQL/Internal/Output.hs. There isn't a way of encoding errors for specific fields, there's just one top-level `errors` key that has all the errors.

We need just need a structured type (or types, I guess) that we can convert to [GraphQL.Internal.Output.Error](https://github.com/jml/graphql-api/blob/master/src/GraphQL/Internal/Output.hs#L74) when we want to serialise it.

tests/TypeApiTests.hs
teh: github started hiding larger diffs, click on "load diff" above ...

tests/TypeApiTests.hs
teh: It's an error because we run over a custom `ExceptT` monad (only in this specific test though)

jml: Sorry, I don't quite understand. I get that we're running over a user-level `ExceptT` in this test, but maybe we should not do that any more, since nothing in the code makes use of it?

src/GraphQL/Server.hs
```diff
@@ -37,18 +38,16 @@ import GraphQL.Internal.Schema (HasName(..))
...
+data QueryError =
+  QueryError Text
+  | InternalError Text
+  | InvalidFieldError Text
+  deriving (Show, Eq)
 
--- | MonadThrow requires an instance of Exception so we create a
--- newtype for GraphQL errors.
-newtype QueryError = QueryError Text deriving (Show, Eq)
 instance Exception QueryError
```
jml: Is this still necessary / desirable?

src/GraphQL/Server.hs
```diff
@@ -77,11 +76,21 @@ infixr 8 :<>
 data a :<|> b = a :<|> b
 infixr 8 :<|>
 
+
+-- Result collects errors and values at the same time unless a handler
+-- tells us to bail out in which case we stop the processing
+-- immediately.
+-- TODO: Make QueryError a Monoid and aggregate according to that.
```
jml: I'm not sure we want to make it a `Monoid`, but it's definitely worth a try.

src/GraphQL/Server.hs
```diff
@@ -77,11 +76,21 @@ infixr 8 :<>
...
+data Result = Result [QueryError] GValue.Value
+ok :: GValue.Value -> Result
+ok v = Result [] v
+
+
 -- TODO instead of SelectionSet we want something like
 -- NormalizedSelectionSet which has query fragments etc. resolved.
-class (MonadThrow m, MonadIO m) => HasGraph m a where
+class (Monad m) => HasGraph m a where
```
jml: You don't need to say `Monad m` here, and I don't think you should. In general, it's a good idea to put constraints where they are actually needed.

The only reason you might want to put it here is for `HasGraph m a` instances to always imply `Monad m` (`Eq a => Ord a` being the canonical example). But we already explicitly say `Monad m` on each instance.

src/GraphQL/Server.hs
```diff
@@ -101,32 +110,39 @@ class ReadValue a where
...
-    let a = traverse (flip (buildResolver @m @hg) selectionSet) handler
-    in map GValue.toValue a
-
-instance forall m ks enum. (MonadThrow m, MonadIO m, GraphQLEnum enum) => HasGraph m (Enum ks enum) where
+  buildResolver handler selectionSet = map resultValues a
+    where
+      a = traverse (flip (buildResolver @m @hg) selectionSet) handler
+      -- Aggregate results. TODO: Result should probably be a monoid
+      -- so we can collect values but I'm not sure what the monoid of
+      -- Value itself is.
```
jml: Me either :)

Worth thinking about though.

src/GraphQL/Server.hs
```diff
@@ -101,32 +110,39 @@ class ReadValue a where
...
+      resultValues :: [Result] -> Result
+      resultValues r =
+        let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
+        in Result errs (GValue.toValue valueList)
```
jml: I think this should probably be a top-level function that lives next to the definition of `Result`.

src/GraphQL/Server.hs
```diff
@@ -77,11 +76,21 @@ infixr 8 :<>
...
 infixr 8 :<|>
 
+
+-- Result collects errors and values at the same time unless a handler
+-- tells us to bail out in which case we stop the processing
+-- immediately.
+-- TODO: Make QueryError a Monoid and aggregate according to that.
+data Result = Result [QueryError] GValue.Value
+ok :: GValue.Value -> Result
+ok v = Result [] v
```
jml: You should probably make `Result` implement `Functor`, `Applicative`, and `Monad`. Then `ok` becomes `pure`. We can still keep around in its specific form, but it'll make a bunch of the code below simpler (I think).

src/GraphQL/Server.hs
```diff
@@ -180,27 +196,37 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
...
+-- Iterate through handlers (zipped together with their type
+-- definition) and execute handler if the name matches.
+--
+-- TODO: tuple return not great (but either is wrong, too because it discards errors.
+type ResolveFieldResult = ([QueryError], Maybe GValue.ObjectField)
+resolveField :: forall a (m :: Type -> Type). BuildFieldResolver m a
+  => FieldHandler m a -> m ResolveFieldResult -> AST.Selection -> m ResolveFieldResult
+resolveField handler nextHandler selection@(AST.SelectionField (AST.Field _ queryFieldName _ _ _)) =
+  case buildFieldResolver @m @a handler selection of
+    -- TODO the fact that this doesn't fit together nicely makes me think that ObjectField is not a good idea)
```
jml: Interesting. Worth exploring later, I think.

src/GraphQL/Server.hs
```diff
@@ -180,27 +196,37 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
...
-resolveField _ _ f = queryError ("Unexpected Selection value. Is the query normalized?: " <> show f)
+data NamedValueResolver m = NamedValueResolver AST.Name (m Result)
+
+
+
+-- Iterate through handlers (zipped together with their type
+-- definition) and execute handler if the name matches.
+--
+-- TODO: tuple return not great (but either is wrong, too because it discards errors.
+type ResolveFieldResult = ([QueryError], Maybe GValue.ObjectField)
```
jml: Re the TODO. Based on the comment on `Result` ("Result collects errors and values at the same time"), `Result` itself should have this ability. You can see this reflected in the structure of `Response` (which should probably _not_ be used for individual field results, since it's intended to represent the whole response). 

I wonder if it would be better just to go for that and remove this type alias. Doing so would at least move away from the anonymous product type and go to an explicit one.

src/GraphQL/Server.hs
```diff
@@ -180,27 +196,37 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
...
+          Result errs value <- resolver
+          pure (errs, Just (GValue.ObjectField queryFieldName value))
+      | otherwise = nextHandler
+resolveField _ _ f =
+  -- TODO, I'm not sure whether this should be a panic instead? (If we
+  -- panic we can avoid the Maybe in the ResolveFieldResult type).
+  --
+  -- An alternative intepretation is that we can formulate
+  -- syntactically valid queries that are still semantically broken in
+  -- which case we should not panic.
```
jml: Basically this last one.

src/GraphQL/Server.hs
```diff
@@ -37,18 +38,16 @@ import GraphQL.Internal.Schema (HasName(..))
 import GraphQL.API
 import GraphQL.Internal.Input (CanonicalQuery)
 
-import Control.Monad.Catch (MonadThrow, throwM, Exception)
+import Control.Monad.Catch (Exception)
+
+data QueryError =
```
jml: I'd like us to rename this type now that it represents more than just query errors. I think `Error` would be a fine name, although `GraphQLError` would also be ok.

src/GraphQL/Server.hs
```diff
@@ -37,18 +38,16 @@ import GraphQL.Internal.Schema (HasName(..))
...
+  QueryError Text
+  | InternalError Text
```
jml: Can we please rename this `SchemaError` (or add something to the sum type to specifically indicate errors in the definition of the schema)?

src/GraphQL/Server.hs
```diff
@@ -37,18 +38,16 @@ import GraphQL.Internal.Schema (HasName(..))
...
+  | InvalidFieldError Text
```
jml: We talked earlier about having a hard distinction between errors that spring from users (client-side errors, 400s, query errors) and errors that spring from the server code (500s). Which is this?

src/GraphQL/Server.hs
```diff
@@ -180,27 +196,37 @@ instance forall v. ReadValue v => ReadValue (Maybe v) where
...
+          pure (errs, Just (GValue.ObjectField queryFieldName value))
+      | otherwise = nextHandler
+resolveField _ _ f =
+  -- TODO, I'm not sure whether this should be a panic instead? (If we
+  -- panic we can avoid the Maybe in the ResolveFieldResult type).
+  --
+  -- An alternative intepretation is that we can formulate
+  -- syntactically valid queries that are still semantically broken in
+  -- which case we should not panic.
+  pure ([InternalError ("Unexpected Selection value. Is the query normalized?: " <> show f)], Nothing)
```
jml: This can maybe wait for a second PR, but I'd like to see us generating more structured errors in this code (e.g. `SchemaError (UnexpectedSelection f)`) and then using something like `formatNameError` to generate user-friendly messages. Probably we'd have a typeclass with a `formatError` method, or something.

The reason is that it helps to reason about all the possible things that can go wrong.

If we made this change, we'd see that `QueryError` (or `Error`, if we rename it such), would have very very few `Text` fields.

src/GraphQL/Server.hs
```diff
@@ -247,53 +272,52 @@ type family RunFieldsType (m :: Type -> Type) (a :: [Type]) = (r :: Type) where
...
+         , Monad m
          ) => RunFields m (f :<> fs) where
-  type RunFieldsHandler m (f :<> fs) = FieldHandler m f :<> RunFieldsHandler m fs
-  runFields (lh :<> rh) selection =
-    resolveField @f @m fallback lh selection
+  runFields (handler :<> nextHandlers) selection =
+    resolveField @f @m handler nextHandler selection
     where
-      fallback = runFields @m @fs rh selection
+      nextHandler = runFields @m @fs nextHandlers selection
```
jml: That is definitely a better name. Thank you.

src/GraphQL/Server.hs
jml: If necessary, we can leave a branch of the sum type to mean "once we do query validation properly, these should probably go away".

src/GraphQL/Server.hs
teh: Removed todo, I'm not sure we can easily do it.

src/GraphQL/Server.hs
teh: removed, thanks

src/GraphQL/Server.hs
teh: done

src/GraphQL/Server.hs
teh: I've changed this to the parametrized result but I think it gives a false sense of generalization - `ObjectField` is so different from `Value` that we can't even have a generic aggregator. IMO it'd be better to stick with some module-local `ResolveFieldResult` type to avoid giving the impression that it's anything like result. What do you think?

src/GraphQL/Server.hs
teh: removed comment

src/GraphQL/Server.hs
teh: +1, this PR already introduces more structure than before, but even more would be better

src/GraphQL/Server.hs
```diff
@@ -76,11 +73,34 @@ infixr 8 :<>
...
+aggregateResults :: [Result GValue.Value] -> Result GValue.Value
+aggregateResults r =
+  let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
+  in Result errs (GValue.toValue valueList)
+
+
+instance Applicative Result where
+  pure v = Result [] v
+  -- TODO it's not obvious to me that this is the right function...
+  (Result e f) <*> (Result ea ev) = Result (e <> ea) (f ev)
```
jml: Looks right to me. We always apply the function (whence 'applicative' gets its terrible name) and our context always appends errors. 

If in doubt, write some property tests for the applicative laws.

As a style thing, I'd say `e1` and `e2`, and change `ev` to `x`, but that's mostly personal preference.

src/GraphQL/Server.hs
```diff
@@ -76,11 +73,34 @@ infixr 8 :<>
...
+
+ok :: GValue.Value -> Result GValue.Value
+ok = pure
```
jml: Personally I'd just ditch this, but no big deal.

src/GraphQL/Server.hs
```diff
@@ -112,34 +132,40 @@ instance Defaultable Bool
...
 
+instance Defaultable (Maybe a) where
+  -- | The default for @Maybe a@ is @Nothing@.
+  defaultFor _ = pure Nothing
+
 -- TODO not super hot on individual values having to be instances of
 -- HasGraph but not sure how else we can nest either types or
 -- (Object _ _ fields). Maybe instead of field we need a "SubObject"?
-instance forall m. (MonadThrow m, MonadIO m) => HasGraph m Int32 where
+instance forall m. (Monad m) => HasGraph m Int32 where
```
jml: Pretty sure you don't need `Monad` constraint here, or anywhere for the scalar types.

src/GraphQL/Server.hs
```diff
@@ -193,34 +222,33 @@ type family FieldHandler m (a :: Type) :: Type where
   FieldHandler m (Field ks t) = Handler m t
   FieldHandler m (Argument ks t :> f) = t -> FieldHandler m f
 
-class (MonadThrow m, MonadIO m) => BuildFieldResolver m a where
-  buildFieldResolver :: FieldHandler m a -> AST.Selection -> Either QueryError (NamedFieldExecutor m)
+class (Monad m) => BuildFieldResolver m a where
```
jml: This doesn't need to have the `Monad m` class constraint.

src/GraphQL/Server.hs
jml: Although I see later that it helps avoid `pure . pure` chains. Fair enough.

src/GraphQL/Server.hs
```diff
@@ -76,11 +73,34 @@ infixr 8 :<>
...
+-- tells us to bail out in which case we stop the processing
+-- immediately.
+data Result a = Result [ResolverError] a deriving (Show, Functor)
+
+-- Aggregating results keeps all errors and creates a ValueList
+-- containing the individual values.
+aggregateResults :: [Result GValue.Value] -> Result GValue.Value
+aggregateResults r =
+  let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
+  in Result errs (GValue.toValue valueList)
```
jml: This is just:

```haskell
GValue.toValue <$> sequenceA r
```

Which I figured out from the following logic:

```haskell
[Result GValue.Value] -> Result GValue.Value -- starting type
[Result a] -> Result a -- abstract 'Value'
[Result a] -> Result [a] -- replace second 'a' with list, because that's what it really is (Value masks things)
Applicative f => [f a] -> f [a] -- abstract 'Result' to 'Applicative'
(Applicative f, Traversable t) => t (f a) -> f (t a) -- abstract list to traversable
```

The logic might be wrong, but it's also what I'd expect to fall out of your very natural definition of applicative below.

src/GraphQL/Server.hs
```diff
@@ -164,27 +190,30 @@ lookupValue name args = case find (\(AST.Argument name' _) -> name' == name) arg
...
+  case buildFieldResolver @m @a handler selection of
+    -- TODO the fact that this doesn't fit together nicely makes me think that ObjectField is not a good idea)
+    Left err -> pure (Result [err] (Just (GValue.ObjectField queryFieldName GValue.ValueNull)))
+    Right (NamedValueResolver name' resolver) -> runResolver name' resolver
+  where
+    runResolver :: AST.Name -> m (Result GValue.Value) -> m ResolveFieldResult
+    runResolver name' resolver
+      | queryFieldName == name' = do
+          Result errs value <- resolver
+          pure (Result errs (Just (GValue.ObjectField queryFieldName value)))
```
jml: The unpacking & reconstruction might perhaps give readability gains, but this could also be written as:

```haskell
Just . GValue.ObjectField queryFieldName <$>  resolver
```

src/GraphQL/Server.hs
```diff
@@ -287,94 +314,9 @@ instance forall typeName interfaces fields m.
...
-
-  -- query "{ ... on Human { name } }"
-  -- [SelectionInlineFragment (InlineFragment (NamedType "Human") [] [SelectionField (Field "" "name" [] [] [])])]
-  buildResolver handler selection = do
-    -- GraphQL invariant is that all items in a Union must be objects
-    -- which means 1) they have fields 2) They are ValueMap
-    values <- map GValue.unionObjects (traverse (runUnion @m @(RunUnionType m ru) handler) selection)
-    maybe (panic $ "Duplicate fields in values: " <> show values) (pure . GValue.ValueObject) values
+    r <- forM selectionSet (runFields @m @(RunFieldsType m fields) handler)
+    let (errs, fields) = foldr' (\(Result ea fa) (eb, fbs) -> (eb <> ea, fa:fbs)) ([], []) r
```
jml: As noted above, I think this so close to `sequenceA` you should probably use it.

```haskell
let (Result errs obj)  = GValue.makeObject . catMaybes <$> sequenceA r
pure $ case obj of
  Nothing -> ...
  Just object -> ...
```

src/GraphQL/Server.hs
jml: Looking at the code, I still think the generalisation is worth pursuing.

Another reason is for things like having the selection set resolver return `Result Object` _not_ `Result Value`.

jml:
Just occurred to me that this isn't miles off the [Writer monad](http://dev.stephendiehl.com/hask/#writer-monad).

src/GraphQL/Server.hs
teh: quickcheck says no, it's not the same (though I can't immediately say why not):

```haskell
aggregateResults :: [Result GValue.Value] -> Result GValue.Value
aggregateResults r = GValue.toValue <$> sequenceA r

aggregateResults2 :: [Result GValue.Value] -> Result GValue.Value
aggregateResults2 r =
  let (errs, valueList) = foldl' (\(eb, vb) (Result ea va) -> ((eb <> ea), va:vb)) ([], []) r
  in Result errs (GValue.toValue valueList)

Q.quickCheck $ Q.verbose $ \x -> aggregateResults (map pure x) == aggregateResults2 (map pure x)
```

```haskell
Falsifiable (after 4 tests and 1 shrink):     
[ValueList (List [ValueObject (Object {objectFields = [ObjectField (Name {getNameText = "Pgq"}) (ValueInt 2)]}),ValueObject (Object {objectFields = []}),ValueObject (Object {objectFields = []})]),ValueEnum (Name {getNameText = "itFz"})]
```

src/GraphQL/Server.hs
teh: changed style

src/GraphQL/Server.hs
teh: also constrains types which results in slightly nicer errors that help me debug but we can remove when we're done developing

src/GraphQL/Server.hs
teh: relaxed constraints to Functor and Applicative

src/GraphQL/Server.hs
teh: Hm, fmapping like that misses the Monad `m` level, no? I'm sure I could double map but I'm not convinced it helps readability. Feel free to update if you prefer the double map style

src/GraphQL/Server.hs
teh: moved constraints closer to where they are needed

src/GraphQL/Server.hs
teh: I'll let you fix that - quickcheck failed for the `aggregateResults` case so I don' trust this change either. I need to work my way through this slowly.

src/GraphQL/Server.hs
jml: I checked this out and it's because `sequenceA` is doing the right thing whereas the custom fold is reversing the list.

jml:
```haskell
*GraphQL.Server Protolude GraphQL.Internal.AST Data.Maybe
λ> let y = [GValue.ValueEnum (unsafeMakeName "first"), GValue.ValueEnum (unsafeMakeName "second")]
*GraphQL.Server Protolude GraphQL.Internal.AST Data.Maybe| 
*GraphQL.Server Protolude GraphQL.Internal.AST Data.Maybe
λ> sequenceA (map ok y)
Result [] [ValueEnum (Name {getNameText = "first"}),ValueEnum (Name {getNameText = "second"})]
*GraphQL.Server Protolude GraphQL.Internal.AST Data.Maybe
λ> aggregateResults2 (map ok y)
Result [] (ValueList (List [ValueEnum (Name {getNameText = "second"}),ValueEnum (Name {getNameText = "first"})]))
```

src/GraphQL/Server.hs
jml: Nope. You don't do any branching on the result of resolver.

```
:t resolver ==> m (Result e a)
:t pure (Result err (Just (ObjectField queryFieldName value))) ==> m (Result e (Maybe ObjectField))
```

Classic functor territory, no double-map needed.

src/GraphQL/Server.hs
jml: As pointed out in chat, this is buggy and the sequence version is not. However, I'm always happy to eliminate inessential complexity in follow-up.

jml: Consequence of this is that this instance can just be `Functor`.

teh: I tried and I don't think we can. Maybe I can relax some of the other constraints to make it work but I couldn't make it work.

teh:
Agree could technically be a `WriterT` but IMO the direct collection or errors and results is more readable.

jml:
> Agree could technically be a WriterT but IMO the direct collection or errors and results is more readable.

Did you try this and discover that you preferred the existing way, or is this a guess? If the latter, would you mind if I made an attempt at some later point?

PR #44, opened 2016-12-31
Extract FromValue from ReadValue

jml:
We talked about this earlier, so hopefully this will be uncontroversial:

* Split `ReadValue` typeclass into one that converts values into Haskell objects (`FromValue`) and one that provides defaults (`Defaultable`)
* Change `FromValue` to operate on our "literal" values (which won't have variables, and won't have duplicate fields), rather than AST values (which can be anything that's syntactically valid)
* Change the interface for `Defaultable` to be `Maybe` rather than `Either`, since I figure the error message will always be the same, and users won't want to customise it.
* Roundtripping methods & tests 

If there are conflicts with your error branch, I'll resolve them here.

### DIFF
--- a/src/GraphQL/Server.hs
+++ b/src/GraphQL/Server.hs
@@ -83,17 +83,6 @@ class (MonadThrow m, MonadIO m) => HasGraph m a where
   buildResolver :: Handler m a -> CanonicalQuery -> m GValue.Value
 
 
--- | @a@ can be converted from a GraphQL 'Value' to a Haskell value.
---
--- The @FromValue@ instance converts 'AST.Value' to the type expected by the
--- handler function. It is the boundary between incoming data and your custom
--- application Haskell types.
-class FromValue a where
-  -- TODO: Rename to fromValue.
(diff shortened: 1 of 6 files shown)


src/GraphQL/Server.hs
```diff
@@ -83,20 +82,34 @@ class (MonadThrow m, MonadIO m) => HasGraph m a where
...
+-- To specify a default, implement this typeclass.
+--
+-- The default implementation is to say that there *is* no default for this
+-- type.
+class Defaultable a where
+  -- | defaultFor returns the value to be used when no value has been given.
+  defaultFor :: AST.Name -> Maybe a
+  defaultFor _ = empty
+
+-- | Called when we're missing a value of type @a@ with the given 'AST.Name'.
```
teh: Maybe something like "Called when the schema expects an input argument @name@ but @name@ has not been provided in the query" ?

jml: Changed, thanks.

PR #45, opened 2016-12-31
Anonymous queries

jml:
In #42, @teh was surprised about the behaviour of node names. This is partly due to the structure of the AST, which previous just said that an anonymous query was a query with an empty name. 

This PR gives anonymous queries a distinct branch in the AST, updates the validation logic to be a bit more explicit, and fixes the parser while it's at it (it more or less falls out from having correct types). 

Also changes the top level "document" concept, splitting it into query document and schema document. The reasoning here is that you just can't send type definitions in a query, so don't bother trying to parse them while parsing queries. The upshot is that it further simplifies validation.


jml:
Awesome, thanks. I've been building on this to do "proper" validation. Will be a big branch. Stay tuned.

Issue #21, opened 2016-12-16
"Input" API for parsing, validating and normalizing graphql queries


teh:
The following bit of code:

```
import qualified Data.GraphQL.AST as AST
import Data.GraphQL.Parser (document)
import GraphQL.Value (Value)
import Data.Attoparsec.Text (parseOnly, endOfInput)

query :: Text -> AST.SelectionSet
query q =
  let Right (AST.Document [AST.DefinitionOperation (AST.Query (AST.Node _ _ _ selectionSet))]) =
       parseOnly (document <* endOfInput) q
  in selectionSet
```

could got into a function that returns a `CanonicalQuery` (whose type we can refine at a later stage).

jml:
Going to track the rest of the bits of this in more specific tickets.

PR #46, opened 2017-01-02
Implement most of query validation

jml:
The `Validation` module now contains a structured `QueryDocument` data type ready for use by the executor.

Scope of the module is defined by the [Validation](https://facebook.github.io/graphql/#sec-Validation) chapter of the spec, except as otherwise noted.

Over-arching philosophy is to do "constructive" validation, making new objects that are the ones we wish we had, and reporting errors where not found. Thus, we use Maps to show that names are unique.

There is still more to do, but this is big enough and complete enough that it should be reviewed & merged.

It will probably be easier to review the new file than to review the diff.

### DIFF
--- a/src/GraphQL/Internal/Validation.hs
+++ b/src/GraphQL/Internal/Validation.hs
@@ -1,104 +1,416 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+-- | Transform GraphQL query documents from AST into valid structures
+--
+-- This corresponds roughly to the
+-- [Validation](https://facebook.github.io/graphql/#sec-Validation) section of
+-- the specification, except where noted.
+--
+-- One core difference is that this module doesn't attempt to do any
+-- type-level validation, as we attempt to defer all of that to the Haskell
+-- type checker.
+--
+-- Still missing:
+--
+--   * variable validation <https://facebook.github.io/graphql/#sec-Validation.Variables>
+--   * field selection merging <https://facebook.github.io/graphql/#sec-Field-Selection-Merging>
+--   * input object field uniqueness <https://facebook.github.io/graphql/#sec-Values>
+--
+-- Deliberately not going to do:
+--
+--   * field selections on compound types <https://facebook.github.io/graphql/#sec-Field-Selections-on-Objects-Interfaces-and-Unions-Types>
+--   * leaf field selections <https://facebook.github.io/graphql/#sec-Leaf-Field-Selections>
+--   * argument names <https://facebook.github.io/graphql/#sec-Argument-Names>
+--   * argument value type correctness <https://facebook.github.io/graphql/#sec-Argument-Values-Type-Correctness>
+--   * fragment spread type existence <https://facebook.github.io/graphql/#sec-Fragment-Spread-Type-Existence>
+--   * fragments on compound types <https://facebook.github.io/graphql/#sec-Fragments-On-Composite-Types>
+--   * fragment spread is possible <https://facebook.github.io/graphql/#sec-Fragment-spread-is-possible>
+--   * directives are defined <https://facebook.github.io/graphql/#sec-Directives-Are-Defined>
+--   * directives are in valid locations <https://facebook.github.io/graphql/#sec-Directives-Are-In-Valid-Locations>
+--   * variable default values are correctly typed <https://facebook.github.io/graphql/#sec-Variable-Default-Values-Are-Correctly-Typed>
+--   * variables are input types <https://facebook.github.io/graphql/#sec-Variables-Are-Input-Types>
+--   * all variable usages are allowed <https://facebook.github.io/graphql/#sec-All-Variable-Usages-are-Allowed>
+--
+-- Because all of the above rely on type checking.
 module GraphQL.Internal.Validation
   ( ValidationError(..)
-  , ValidDocument
+  , QueryDocument
   , validate
   , getErrors
+  -- * Operating on validated documents
+  , getOperation
   -- * Exported for testing
   , findDuplicates
   ) where
 
 import Protolude
 
+import Data.List.NonEmpty (NonEmpty(..))
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Map as Map
+import qualified Data.Set as Set
 import qualified GraphQL.Internal.AST as AST
+-- Directly import things from the AST that do not need validation, so that
+-- @AST.Foo@ in a type signature implies that something hasn't been validated.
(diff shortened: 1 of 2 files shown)


src/GraphQL/Internal/Validation.hs
```diff
@@ -1,104 +1,416 @@
...
+-- that make up fragment definitions (see 'validateFragmentDefinitions'). At
+-- this stage, @spread@ will be 'UnresolvedFragmentSpread'.
+--
+-- Once we have a known-good map of fragment definitions, we can do the next
+-- phase of validation, which checks that references to fragments exist, that
+-- all fragments are used, and that we don't have circular references.
+--
+-- This is encoded as a type variable because we want to provide evidence that
+-- references in fragment spreads can be resolved, and what better way to do
+-- so than including the resolved fragment in the type. Thus, @spread will be
```
teh: missing `@` after spread?

src/GraphQL/Internal/Validation.hs
```diff
@@ -1,104 +1,416 @@
...
+-- Note that this is essentially a definition of 'Traversable' for
+-- 'Selection'. However, we probably also want to have other kinds of
+-- traversals (e.g. for transforming values), so best not to bless one kind
+-- with a type class.
+traverseFragmentSpreads :: Applicative f => (a -> f b) -> Selection a -> f (Selection b)
+traverseFragmentSpreads f selection =
+  case selection of
+    SelectionField (Field alias name args directives ss) ->
+      SelectionField <$> (Field alias name args directives <$> childSegments ss)
+    SelectionFragmentSpread x -> SelectionFragmentSpread <$> f x
```
teh: optional nit: maybe move `SelectionFragmentSpread <$> f x` to next line for regularity with the other lines

src/GraphQL/Internal/Validation.hs
```diff
@@ -1,104 +1,416 @@
...
+      SelectionField <$> (Field alias name <$> validateArguments args <*> validateDirectives directives <*> childSegments ss)
+    AST.SelectionFragmentSpread (AST.FragmentSpread name directives) ->
+      SelectionFragmentSpread <$> (UnresolvedFragmentSpread name <$> validateDirectives directives)
+    AST.SelectionInlineFragment (AST.InlineFragment typeCond directives ss) ->
+      SelectionInlineFragment <$> (InlineFragment typeCond <$> validateDirectives directives <*> childSegments ss)
+  where
+    childSegments = traverse validateSelection
+
+-- | Resolve the fragment references in a selection, accumulating a set of
+-- visited fragment names.
```
teh: what does "visited" mean in this context?

src/GraphQL/Internal/Validation.hs
```diff
@@ -112,3 +424,59 @@ findDuplicates xs = findDups (sort xs)
...
+throwErrors :: NonEmpty e -> Validator e a
+throwErrors = Validator . Left
+
+-- | Map over each individual error on a validation. Useful for composing
+-- validations.
+--
+-- This is /somewhat/ like 'first', but 'Validator' is not, and cannot be, a
+-- 'Bifunctor', because the left-hand side is specialized to @NonEmpty e@,
+-- rather than plain @e@. Also, whatever function were passed to 'first' would
+-- get the whole non-empty list, whereas 'mapErrors' works on one element at a
```
teh: Thanks, fantastic doc string! (not actionable)

src/GraphQL/Internal/Validation.hs
```diff
@@ -112,3 +424,59 @@ findDuplicates xs = findDups (sort xs)
...
+-- be composed, and ensures that *all* validation errors bubble up.
+instance Applicative (Validator e) where
+  pure x = Validator (Right x)
+  Validator (Left e1) <*> (Validator (Left e2)) = Validator (Left (e1 <> e2))
+  Validator (Left e) <*> _ = Validator (Left e)
+  Validator _ <*> (Validator (Left e)) = Validator (Left e)
+  Validator (Right f) <*> Validator (Right x) = Validator (Right (f x))
+
+-- | The monad on Validator stops processing once errors have been raised.
+instance Monad (Validator e) where
```
teh: Can you derive Monad? I.e.

```haskell
newtype Validator e a = Validator { runValidator :: Either (NonEmpty e) a } deriving (Eq, Show, Functor, Monad)
```

src/GraphQL/Internal/Validation.hs
jml: Yep. Fixed.

src/GraphQL/Internal/Validation.hs
jml: Done.

src/GraphQL/Internal/Validation.hs
jml: We traverse through the tree of fragment references, "visiting" each and then resolving them. It's language borrowed from the spec, which I also found confusing in context.

Have changed the comment to:

```haskell
-- | Resolve the fragment references in a selection, accumulating a set of
-- the fragment names that we have resolved.
--
-- We're doing a standard depth-first traversal of fragment references, where
-- references are by name, so the set of names can be thought of as a record
-- of visited references.
```

src/GraphQL/Internal/Validation.hs
jml: Yep. Have changed to do so. I'm not sure it does the same thing, but I figure the burden is on me to prove that, since it means more code.

src/GraphQL/Internal/Validation.hs
```diff
@@ -1,104 +1,416 @@
...
+import qualified Data.Set as Set
 import qualified GraphQL.Internal.AST as AST
+-- Directly import things from the AST that do not need validation, so that
+-- @AST.Foo@ in a type signature implies that something hasn't been validated.
+import GraphQL.Internal.AST (Name, Alias, TypeCondition)
+
+-- | A valid query document.
+--
+-- Construct this using 'validate' on an 'AST.QueryDocument'.
+data QueryDocument
```
teh: To check: `QueryDocument` is what we are going to use for execution (currently in `Server.hs`)?

I'm mildly surprised that we don't need to change `Server.hs` after changing the structure of QueryDocument.

jml: The difference is "need" vs "want". `Server.hs` operates on AST, and AST is the same as it has been since #45, so we can change this with impunity.

However, we really should be using this stuff, and we can after this PR is done.

jml:
Part of #21

PR #61, opened 2017-01-03
Use the new validation code in the resolver

jml:
Changes a bunch of types of things.

Because I didn't want to duplicate code, I made a tentative approach at a public API, which I put in the top-level GraphQL. I figure it's easy to correct a wrong thing than to make a right thing.


jml:
Follow up to #46

jml:
Actually, I want to do a pass of this before review. Please hold off.

jml:
OK, good to go now.

src/GraphQL/Internal/Validation.hs
```diff
@@ -183,18 +205,42 @@ validateArguments args = mapErrors DuplicateArgument (makeMap [(name, value) | A
...
   deriving (Eq, Show)
 
+instance HasName (Field' spread) where
+  getName (Field' _ name _ _ _) = name
+
+type Field = Field' FragmentSpread
+
+-- | Get the arguments of a field.
+getArguments :: Field -> Arguments
+getArguments (Field' _ _ args _ _) = args
```
jml: Essentially, yes. 

My thinking is that it seems a shame to go to all the work of constructively proven valid objects (and in many cases, the type signature does constitute a proof of validity), and then let just any bozo construct any random thing and pretend it's valid.

Secondarily, the data types are still in flux. A forthcoming PR will change the validation data types to do variable validation and selection set validation, and another will do variable substitution, fragment handling, and directive processing. Providing functions like these give a useful choke point for making those changes quickly without having to reason globally. "What does the resolver want? Oh, it only needs the arguments for this thing. I'll have it ask for those and I'll figure out how to provide them." 

View patterns actually might be a good fit for this module. I looked into them, for what we're doing at the moment, they are lots more typing.

FWIW, the most important data hiding functions are `getFields` and `getFieldSelectionSets`, which control _traversal_, which is qualitatively different from getting an attribute. (e.g. `getFields` has logic in it, and may well continue to do so).

src/GraphQL/Resolver.hs
```diff
@@ -251,14 +252,14 @@ instance forall ks t f m.
   , HasAnnotatedInputType t
   , Monad m
   ) => BuildFieldResolver m (API.Argument ks t :> f) where
-  buildFieldResolver handler selection@(AST.SelectionField (AST.Field _ _ arguments _ _)) = do
+  buildFieldResolver handler field = do
     argument <- first SchemaError (API.getArgumentDefinition @(API.Argument ks t))
     let argName = getName argument
+    let arguments = getArguments field
```
jml: Just realised that this code should delegate even more work to the data structures themselves.

Rather than saying "field, get my thy arguments" and then "arguments, get me the argument named thus", we should instead say "field, pop this argument".

That would mean:
  1. the resolver could never resolve the same value twice
  2. half as much time spent on lookups

So, something like:

```haskell
let argName = getName argument
(field', value) <- case popArgument field (defaultFor argName) of
  Nothing -> throwError (ValueMissing argName)
  Just (field', v) -> do
    v' <- first (InvalidValue argName) (GValue.fromValue @t v)
    pure (field', v)
buildFieldResolver @m @f (handler value) field'
```

Where,

```haskell
popArgument :: Field -> Name -> Maybe Value -> Maybe Value
popArgument field name default =
  case lookup name (getArguments field) <|> default of
    Nothing -> Nothing
    Just value -> Just (deleteArgument field name, value)
  where
    deleteArgument = notImplemented
```

jml: As I might have mentioned somewhere, I'm working on variable validation atm, so this change could go in there or in this PR.

teh: Popping "used" arguments is a good idea!

src/GraphQL/Internal/Validation.hs
```diff
@@ -64,15 +74,27 @@ data QueryDocument
...
+getSelectionSet (Query _ _ ss) = ss
+getSelectionSet (Mutation _ _ ss) = ss
+
 -- | Type alias for 'Query' and 'Mutation' constructors of 'Operation'.
-type OperationType = [AST.VariableDefinition] -> Directives -> [Selection FragmentSpread] -> Operation
+type OperationType = [AST.VariableDefinition] -> Directives -> SelectionSet -> Operation
 
 newtype Operations = Operations (Map Name Operation) deriving (Eq, Show)
 
+type SelectionSet = [Selection]
```
teh: `SelectionSet` is actually ordered so the name isn't super great (I know the name is from the spec so let's keep it, just something I noticed)

src/GraphQL/Resolver.hs
```diff
@@ -105,11 +113,11 @@ ok :: GValue.Value -> Result GValue.Value
 ok = pure
 
 
--- TODO instead of SelectionSet we want something like
--- NormalizedSelectionSet which has query fragments etc. resolved.
+-- TODO: It's not crystal clear that "SelectionSet" should be the entry point.
```
teh: I think for `HasGraph` it's the correct entry point because we need to be able to nest, and there are no nested `Operation`s other than the top level one.

But yea: I think we need another top-level entry point as well to build the full resolver.

src/GraphQL/Internal/Validation.hs
```diff
@@ -183,18 +205,42 @@ validateArguments args = mapErrors DuplicateArgument (makeMap [(name, value) | A
...
-data Field spread
-  = Field (Maybe Alias) Name Arguments Directives [Selection spread]
+data Field' spread
+  = Field' (Maybe Alias) Name Arguments Directives [Selection' spread]
   deriving (Eq, Show)
 
+instance HasName (Field' spread) where
+  getName (Field' _ name _ _ _) = name
+
+type Field = Field' FragmentSpread
```
teh: I find the `spread` annotations a bit confusing - isn't the primary difference that one `Field` has everything resolved, and the other doesn't? Would `Resolved  | Unresolved` be accurate?

src/GraphQL.hs
```diff
@@ -0,0 +1,55 @@
...
+  -- | Parsed, but failed validation.
+  --
+  -- See <https://facebook.github.io/graphql/#sec-Validation> for more
+  -- details.
+  | ValidationError ValidationErrors
+  deriving (Eq, Show)
+
+-- | Turn some text into a valid query document.
+--
+-- TODO: jml hates the name of this function, \"process\" is so meaningless.
```
teh: `compile` maybe? Like regexes or progams are compiled?

src/GraphQL/Internal/Validation.hs
jml: Yes, that's the difference. `Resolved | Unresolved` would be accurate, but not precise. More accurate would be "Field that has resolved fragment spreads". So `ResolvedFragmentSpread | UnresolvedFragmentSpread` (although note that these are different types, and you can't normally write `a | b` in Haskell). Because this module follows a convention of calling fully resolved things by their simplest name, it's `FragmentSpread | UnresolvedFragmentSpread`.

FWIW, I did write a large [comment explaining this](https://github.com/jml/graphql-api/blob/master/src/GraphQL/Internal/Validation.hs#L164-L183). If it's not helpful, please let me know how it could be more helpful.

src/GraphQL/Resolver.hs
jml: Agree. Should I ditch / update this comment?

src/GraphQL.hs
jml: Yeah, I like that. Will change.

src/GraphQL/Internal/Validation.hs
teh: Ah yea I remember the comment but I'm not good with "whole module context", I'm a very local reader, that's why I often pick long variable names :) - it's fine as it is, thanks for explaining.

src/GraphQL/Resolver.hs
teh: Maybe, yea. I went through a similar thought process as you when I wrote the following comment: 

```
-- TODO not super hot on individual values having to be instances of
-- HasGraph but not sure how else we can nest either types or
-- (Object _ _ fields). Maybe instead of field we need a "SubObject"?
```

But the current architecture is the only one I can think of that allows arbitrary nesting in a nice way, and nice composing is (to me) a good indicator that the design is OK, even if not super intuitive.

A counter-example would be using `Result` in `resolveField` - I think I used a hammer to push a square peg through a round hole there, and it doesn't quite work.

teh: (feel free to delete that comment as well)

src/GraphQL/Internal/Validation.hs
jml: Suggestions for better names welcome. The key thing is that it's the type of the fragment spread used by the field, and that the reason these types vary (i.e. why they are type _vari_ables) is that we've got some state machines and this variable represents the current state.

teh: It's cool. (We could use DataKinds to elevate a sum type to tag the state manchine states, which might make it slightly clearer what types of `Field` we can have, but again: not blocking this branch, merge when you are happy to merge)

jml: I'd like to see the attempt (but after I've done variable validation & processing).

src/GraphQL/Resolver.hs
jml: Done.

I think the issue w/ `Result` is that it's not yet quite the right kind of abstraction. I have Ideas.

jml:
> I don't have a good intuition on whether the information-hiding changes make it nicer yet, I'll have to actually use these changes in anger.

Cool. Would like to get to a point where using functions to access members is uncontroversial.

PR #63, opened 2017-01-04
Implement type-safe, tagged unions for graphql Unions. 

teh:


src/GraphQL/Resolver.hs
```diff
@@ -326,3 +330,98 @@ instance forall typeName interfaces fields m.
     case obj of
       Nothing -> pure (Result [DuplicateFields r] GValue.ValueNull)
       Just object -> pure (Result errs (GValue.ValueObject object))
+
+
+-- TODO(tom): we're getting to a point where it might make sense to
+-- split resolver into submodules (GraphQL.Resolver.Union  etc.)
```
jml: +1. Or even GraphQL.Internal.Resolver.Union, and have GraphQL.Resolver just re-export the public facing bits.

jml: I want to move AST, Encoder, and Parser under GraphQL.Internal.Syntax, for example.

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +330,98 @@ instance forall typeName interfaces fields m.
...
+unionValue x =
+  DynamicUnionValue (symbolText @name) (unsafeCoerce x)
+
+extractUnionValue ::
+  forall (o :: Type) (union :: Type) m (name :: Symbol) i f.
+  (Monad m, API.Object name i f ~ o, KnownSymbol name) -- TODO probably don't need all constraints here
+  => DynamicUnionValue union m -> Maybe (TypeIndex m o union)
+extractUnionValue (DynamicUnionValue uName uValue) =
+  if uName == symbolText @name
+  then Just (unsafeCoerce uValue)
```
jml: This'll probably need some commentary :)

teh: sorry yea, it's coming - I wasn't going to write comments until I knew the scheme worked.

tests/Examples/UnionExample.hs
```diff
@@ -5,38 +5,33 @@ module Examples.UnionExample  where
...
 o1 = pure (pure "hello from O1")
 
 o2 :: Handler IO O2
-o2 = pure (pure "hello from O2")
+o2 = pure (pure 32)
 
-tHandler :: Handler IO T
-tHandler = o1 :<|> o2
+u1 :: Handler IO U1
+u1 = unionValue @O2 o2
```
teh: `unionValue` is the new external API I'm implementing - the users of our library tag their handler with this. `unionValue` is clever in that o2 needs to match the type specified by `O2`, but the actual handler executed is picked by the inline fragment type name.

teh:
There's one last really important decision, and that's whether to make the union handlers themselves monadic. They are not at the moment.

My intuition says yes, people will sometimes need to run e.g IO actions before deciding which branch of a union to return, but I'm also slightly unhappy about the proliferation of `pure` or `return`. 

If this all becomes a bit much we always have the option of introducing a `PureField` that cannot run any actions, and just takes a value?

teh:
After writing a bit more test code: Yea, we need the extra monadic layer. For usability I folded the `pure` return into the `unionValue` function instead of requiring an extra `pure` before each `unionValue` call. I'd be open to expose a `unionValue'` that's not monadic

tests/Examples/UnionExample.hs
```diff
@@ -1,42 +1,47 @@
...
-tHandler :: Handler IO T
-tHandler = o1 :<|> o2
+exampleQuery :: IO (Result Value)
+exampleQuery = buildResolver @IO @CatOrDog catOrDog (query "{ ... on MiniCat { name meowVolume } ... on MiniDog { barkVolume } }")
 
-exampleQuery :: IO Value
-exampleQuery = buildResolver @IO @T tHandler (query "{ ... on O1 { o1 } ... on O2 { o2 } }")
+-- unionValue can also be used in a list:
+exampleListQuery :: IO (Result Value)
+exampleListQuery = buildResolver @IO @CatOrDogList catOrDogList  (query "{ ... on MiniCat { name meowVolume } ... on MiniDog { barkVolume } }")
```
jml: Example out of these queries would be very nice, but not necessary.

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+  , RunFields m (RunFieldsType m f)
+  , RunUnion m union os
+  ) => RunUnion m union ((API.Object n i f):os) where
+  runUnion duv fragment@(SelectionInlineFragmentPattern _ selection) =
+    case extractUnionValue @(API.Object n i f) @union @m duv of
+      Just handler -> Just <$> (buildResolver @m @(API.Object n i f) handler selection)
+      Nothing -> runUnion @m @union @os duv fragment
+  runUnion _ _ = pure Nothing
+
+-- TODO(tom): AFAICT it should not be possible to ever hit the empty
```
teh: removed todo

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+-- case because the compiler doesn't allow constructing a unionValue
+-- that's not in the Union. If the following code ever gets executed
+-- it's almost certainly a bug in the union code.
+--
+-- We still need to implement this instance for the compiler because
+-- it exhaustively checks all cases when deconstructs the Union.
+instance forall m union. RunUnion m union '[] where
+  runUnion _ _ = notImplemented
```
teh: done

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+
+instance forall m unionName objects.
+  ( Monad m
+  , KnownSymbol unionName
+  , RunUnion m (API.Union unionName objects) objects
+  ) => HasGraph m (API.Union unionName objects) where
+  type Handler m (API.Union unionName objects) = m (DynamicUnionValue (API.Union unionName objects) m)
+  buildResolver mHandler selectionSet = do
+    duv@(DynamicUnionValue _label _) <- mHandler
```
teh: removed

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+    -- we only need to look at the fragment that matches by name:
+    case find (matchFragmentName _label) selectionSet of
+      Nothing -> pure (Result [UnionTypeNotFound] GValue.ValueNull) -- TODO more error detail
```
teh: done

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+      Just fragment -> do
+        -- loop through union handlers and call right one when type matches.
+        r <- runUnion @m @(API.Union unionName objects) @objects duv fragment
+        case r of
+          Just r' -> pure r'
+          Nothing -> pure (Result [ResolveNonUnionType] GValue.ValueNull)
```
teh: probably selectionset as well as this is a query error (done)

teh: This comment no longer applies after `InlineFragment` changes (need a full pattern match now)

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+
+      where
+        matchFragmentName label' (SelectionInlineFragmentPattern (AST.NamedType name') _) =
+          label' == AST.getNameText name'
+        matchFragmentName _ _ = False
+
+
+symbolText :: forall ks. KnownSymbol ks => Text
+symbolText = toS (symbolVal @ks Proxy)
```
teh: It's not the same, the other one is

```
nameFromSymbol :: forall (n :: Symbol). KnownSymbol n => Either NameError Name
nameFromSymbol = makeName (toS (symbolVal @n Proxy))
```

I could write `nameFromSymbol` in terms of `symbolText` but not sure it's worth it

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+
+-- | Translate a @Handler@ into a DynamicUnionValue type required by
+-- @Union@ handlers. This is dynamic, but nevertheless type-safe
+-- because we can only tag with types that are part of the union.
+--
+-- Use e.g. like "unionValue @Cat" if you have an object like this:
+--
+-- > type Cat = Object "Cat" '[] '[Field "name" Text]
```
teh: `unionValue` can't really be used outside of a handler that constrains the 2nd free type argument (the union) + an actual handler, you'll get joyous errors like

```
              • Couldn't match expected type ‘f0 (f1 a0)’
                            with actual type ‘TypeIndex m0 Cat union0’
```

teh: (I added a partial example that compiles + a reference to UnionExample.hs)

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+  => TypeIndex m o union -> m (DynamicUnionValue union m)
+unionValue x =
+  -- TODO(tom) - we might want to move to Typeable `cast` for uValue
+  -- instead of doing our own unsafeCoerce because it comes with
+  -- additional safety guarantees: Typerep is unforgeable, while we
+  -- can still into a bad place by matching on name only. We can't
+  -- actually segfault this because right now we walk the list of
+  -- objects in a union left-to-right so in case of duplicate names we
+  -- only every see one type. That doesn't seen like a great thing to
+  -- rely on though!
```
teh: Added another, more explicit note though I'm not super clear what you are looking for. There's also the comment on `DynamicUnionValue` which says the same thing in a different way.

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+  pure (DynamicUnionValue (symbolText @name) (unsafeCoerce x))
```
teh: For usability, otherwise I have to call `pure (unionValue ..)` everywhere.

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,123 @@ instance forall typeName interfaces fields m.
...
+
+extractUnionValue ::
```
teh: hm - any suggesions? `unionValue` is public API, `extractUnionValue` is internal. That's why I picked the "nice" name for the public API

tests/Examples/UnionExample.hs
```diff
@@ -1,42 +1,47 @@
...
-
-{-
-
-import Protolude hiding (Enum)
-import qualified GraphQL.Internal.AST as AST
-import Data.Attoparsec.Text (parseOnly, endOfInput)
-import GraphQL.Internal.Parser (queryDocument)
-
-
+import Protolude hiding (Enum, U1)
```
teh: removed

tests/Examples/UnionExample.hs
```diff
@@ -1,42 +1,47 @@
...
 
-type T = Union "U" '[O1, O2]
+miniDog :: Handler IO MiniDog
+miniDog = pure (pure 100)
 
-o1 :: Handler IO O1
-o1 = pure (pure "hello from O1")
+catOrDog :: Handler IO CatOrDog
+catOrDog = do
+  name <- pure "Hello" -- we can do nomadic actions
```
teh: fixed

tests/Examples/UnionExample.hs
teh: It's pretty long and looks like this: 

```
Result [] (ValueObject (Object {objectFields = [ObjectField (Name {getNameText = "name"}) (ValueString (String "Hello")),ObjectField (Name {getNameText = "meowVolume"}) (ValueInt 32)]}))
```

and we'll have to keep it up to date by hand - not sure that's a good idea?

jml: Doc test? Actual test?

teh: added a doctest

src/GraphQL/Resolver.hs
```diff
@@ -326,3 +336,130 @@ instance forall typeName interfaces fields m.
...
+  -- Note that unsafeCoerce is safe because we index the type from the
+  -- union with an 'API.Object' whose name we're storing in label. On
+  -- the way out we check that the name is the same, and we know the
+  -- type universe is the same because we annotated DynamicUnionValue
+  -- with the type universe.
+  pure (DynamicUnionValue (symbolText @name) (unsafeCoerce x))
+
+extractUnionValue ::
+  forall (object :: Type) (union :: Type) m (name :: Symbol) interfaces fields.
+  (Monad m, API.Object name interfaces fields ~ object, KnownSymbol name) -- TODO probably don't need all constraints here
```
jml: Is this still a TODO?

PR #64, opened 2017-01-08
Generic enum

teh:
Rough implementation, has some nice properties though not encoded as well as they could be:

* calling `enumValues` fails if there is an invalid name (right now through an incomplete pattern match). `enumValues` is mostly internal so we can decide how we want to fail here. Should tie in with general validation.
* Checks for unary constructors and doesn't compile if it's not. (Try `instance GraphQLEnum A`)

### DIFF
--- a/src/GraphQL/API.hs
+++ b/src/GraphQL/API.hs
@@ -231,7 +231,7 @@ instance forall t. (HasAnnotatedType t) => HasAnnotatedType (List t) where
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedType (Enum ks enum) where
   getAnnotatedType = do
     let name = nameFromSymbol @ks
-    let enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
+    let enums = sequenceA (enumValues @enum) :: Either NameError [Name]
     let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedType . TypeDefinitionEnum <$> et
 
@@ -282,6 +282,6 @@ instance forall t. (HasAnnotatedInputType t) => HasAnnotatedInputType (List t) w
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedInputType (Enum ks enum) where
   getAnnotatedInputType = do
     let name = nameFromSymbol @ks
-        enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
+        enums = sequenceA (enumValues @enum) :: Either NameError [Name]
     let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedInputType . InputTypeDefinitionEnum <$> et
--- a/src/GraphQL/API.hs
+++ b/src/GraphQL/API.hs
@@ -1,8 +1,8 @@
-{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE AllowAmbiguousTypes #-}
-{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeInType #-}
 {-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 -- | Type-level definitions for a GraphQL schema.
 module GraphQL.API
@@ -231,7 +231,7 @@ instance forall t. (HasAnnotatedType t) => HasAnnotatedType (List t) where
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedType (Enum ks enum) where
   getAnnotatedType = do
     let name = nameFromSymbol @ks
-    let enums = sequenceA (enumValues @enum) :: Either NameError [Name]
+    let enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
     let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedType . TypeDefinitionEnum <$> et
 
@@ -282,6 +282,6 @@ instance forall t. (HasAnnotatedInputType t) => HasAnnotatedInputType (List t) w
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedInputType (Enum ks enum) where
   getAnnotatedInputType = do
     let name = nameFromSymbol @ks
-        enums = sequenceA (enumValues @enum) :: Either NameError [Name]
+        enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
     let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedInputType . InputTypeDefinitionEnum <$> et
--- a/src/GraphQL/API.hs
+++ b/src/GraphQL/API.hs
@@ -231,7 +231,8 @@ instance forall t. (HasAnnotatedType t) => HasAnnotatedType (List t) where
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedType (Enum ks enum) where
   getAnnotatedType = do
     let name = nameFromSymbol @ks
-    let et = EnumTypeDefinition <$> name <*> pure (map EnumValueDefinition (enumValues @enum Proxy))
+    let enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
+    let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedType . TypeDefinitionEnum <$> et
 
 instance forall ks as. (KnownSymbol ks, UnionTypeObjectTypeDefinitionList as) => HasAnnotatedType (Union ks as) where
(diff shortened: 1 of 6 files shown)


teh:
Two examples:

``haskell
instance GraphQLEnum B

data B = B1 | B2 | B3 deriving (Generic, Show)
data A = A1 | A2 Text | A3 deriving (Generic, Show)
```

File: src/GraphQL/API.hs [hunk 1]
```diff
@@ -1,8 +1,8 @@
-{-# LANGUAGE TypeFamilies, ScopedTypeVariables, TypeFamilyDependencies #-}
-{-# LANGUAGE GADTs, AllowAmbiguousTypes, UndecidableInstances #-}
-{-# LANGUAGE MultiParamTypeClasses, RankNTypes #-}
-{-# LANGUAGE FlexibleInstances, TypeOperators, TypeInType #-}
-{-# LANGUAGE OverloadedLabels, MagicHash #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE TypeInType #-}
+{-# LANGUAGE TypeOperators #-}
```
jml: Can you please alpha sort these?

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+  ) where
+
+import Protolude hiding (Enum, U1, TypeError)
+import GraphQL.Internal.AST (Name, nameFromSymbol, NameError)
+import qualified GraphQL.Value as GValue
+import GHC.Generics (M1(..), Rep, Meta(..), C1, U1(..), D, (:+:)(..))
+import GHC.TypeLits (KnownSymbol, TypeError, ErrorMessage(..))
+
+
+class GenricEnumValues (f :: Type -> Type) where
```
jml: `GenericEnumValues`

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+{-# LANGUAGE DefaultSignatures #-}
+
+module GraphQL.API.Enum
+  ( GraphQLEnum(..)
+  ) where
+
+import Protolude hiding (Enum, U1, TypeError)
+import GraphQL.Internal.AST (Name, nameFromSymbol, NameError)
+import qualified GraphQL.Value as GValue
+import GHC.Generics (M1(..), Rep, Meta(..), C1, U1(..), D, (:+:)(..))
```
jml: We get many of these from Protolude: https://github.com/sdiehl/protolude/blob/39d56792ced9105e3019841f090bd55f3c12b19f/src/Protolude.hs#L362

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+import GHC.TypeLits (KnownSymbol, TypeError, ErrorMessage(..))
+
+
+class GenricEnumValues (f :: Type -> Type) where
+  genericEnumValues :: [Either NameError Name]
+  genericEnumFromValue :: GValue.Value -> Either Text (f p)
```
jml: Why does this take `Value`? Since the scope of this class is restricted to enums, it should take whatever type `x` is in `ValueEnum x`. 

Then our code that calls `enumToValue` and `enumFromValue` does the pattern match. This renders the "should have been caught at top-level" `panic` redundant. 

… oh, I see that's `Name`. Still `Name` would be better here than `Value` which is overly broad and makes it too easy for implementers to provide buggy implementations. Later (or even in this PR, actually), we can make an `newtype Enum = Enum Name` alias, which would also be useful later for providing that the enum isn't `true`, `false`, or `null`.

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+  genericEnumValues = let name = nameFromSymbol @conName in [name]
+  genericEnumFromValue (GValue.ValueEnum vname) =
+    case nameFromSymbol @conName of
+      Right name -> if name == vname
+                    then Right (M1 U1)
+                    else Left ("Not a valid enum name: " <> show vname)
+      Left x -> Left ("Not a valid enum name: " <> show x)
+  genericEnumFromValue _ = panic "This case should have been caught at top-level. Please file a bug."
+  genericEnumToValue (M1 _) =
+    let Right name = nameFromSymbol @conName
```
jml: Can you please add a specific `panic` for this failed pattern match?

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+  genericEnumFromValue = notImplemented
+  genericEnumToValue = notImplemented
+
+instance forall conName p b sa sb f.
+  ( TypeError ('Text "Constructor not unary: " ':<>: 'Text conName)
+  , KnownSymbol conName
+  ) => GenricEnumValues (C1 ('MetaCons conName p b) (S1 sa sb) :+: f) where
+  genericEnumValues = notImplemented
+  genericEnumFromValue = notImplemented
+  genericEnumToValue = notImplemented
```
jml: Interesting. I've been using `notImplemented` for "not implemented _yet_", whereas you use them here for "won't be implemented". I looked this up and:

- [notImplemented](https://github.com/sdiehl/protolude/blob/4a2ea2735be0f49879f351980a14000f60acce79/docs/Debug.md#notimplemented) — "An undefined expression standing in for a yet to completed program."
- [undefined](https://github.com/sdiehl/protolude/blob/4a2ea2735be0f49879f351980a14000f60acce79/docs/Debug.md#undefined) — "An undefined expression standing in for an incomplete program, unevaluated type witness, or unreachable code branch."

Seems like `undefined` is more appropriate to me.

tests/TypeTests.hs
```diff
@@ -46,7 +46,7 @@ import GraphQL.Value (Value(..))
 data DogCommandEnum = Sit | Down | Heel
 
 instance GraphQLEnum DogCommandEnum where
-  enumValues = map unsafeMakeName ["SIT", "DOWN", "HEEL"]
+  enumValues = map makeName ["SIT", "DOWN", "HEEL"]
```
jml: Want to change one of these examples to use the new Generic stuff?

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+  genericEnumToValue (R1 gv) = genericEnumToValue gv
+
+instance forall conName p b. (KnownSymbol conName) => GenricEnumValues (C1 ('MetaCons conName p b) U1) where
+  genericEnumValues = let name = nameFromSymbol @conName in [name]
+  genericEnumFromValue (GValue.ValueEnum vname) =
+    case nameFromSymbol @conName of
+      Right name -> if name == vname
+                    then Right (M1 U1)
+                    else Left ("Not a valid enum name: " <> show vname)
+      Left x -> Left ("Not a valid enum name: " <> show x)
```
jml: This error gets repeated three times. Please extract a helper function. 

Alternatively, if the type signature change appeals to you, then this becomes the _only_ error, and at that point we might as well make it a `Maybe`, and then generate the nice error text once at the point of use.

src/GraphQL/API/Enum.hs
```diff
@@ -0,0 +1,101 @@
...
+  ) => GenricEnumValues (C1 ('MetaCons conName p b) (S1 sa sb) :+: f) where
+  genericEnumValues = notImplemented
+  genericEnumFromValue = notImplemented
+  genericEnumToValue = notImplemented
+
+
+-- | For each enum type we need 1) a list of all possible values 2) a
+-- way to serialise and 3) deserialise.
+class GraphQLEnum a where
+  enumValues :: [Either NameError Name]
```
jml: This is going to weird people out. 

If we're changing to _this_, we might as well make the interface `[Text]` and call `makeName` ourselves. I don't know how I feel about that. It's more convenient for most cases, but I wonder if we're masking reality in a way that robs users of power.

I think I'd prefer `Text` to `Either NameError Name`, and `Name` to both (but would not insist on `Name` here & now).

src/GraphQL/API.hs
teh: done

src/GraphQL/API/Enum.hs
teh: removed - IMO weird `Protolude` exports those because they are pretty niche but whatever :)

src/GraphQL/API/Enum.hs
teh: fixed

src/GraphQL/API/Enum.hs
teh: good point, updated

src/GraphQL/API/Enum.hs
teh: done

I'm not sure I want to lose the information about the error (e.g. the name that didn't work), so not sure about the Maybe..

src/GraphQL/API/Enum.hs
teh: done

src/GraphQL/API/Enum.hs
teh: Hm turns out I can't generalize this without exporting the NameError constructor, so I kept one of the left constructors. We can maybe make the error a bit more structured, but I'd like to go there when after using this code.

src/GraphQL/API/Enum.hs
teh: Thanks - that's what I had first but it seemed weird. Moved back

src/GraphQL/API/Enum.hs
teh: If we make users use  `makeName` we're leaning on our users to do all the error checking just for constructing the list of `Name` (which can be done statically most of the time). They may then be tempted to use `unsafeMakeName` to avoid the extra code, and then we have to do error checking in `API` because we can't rely on the `Name` contract.

Not sure what the right solution here is but making all users write sth like 

```haskell
enumValues = let Right x = sequenceA [makeName "A", makeName "B"] in x
```

doesn't seem great either..

not changed for now, any other ideas?

tests/TypeTests.hs
teh: Done, but had to re-introduce a Proxy hack because GHC isn't clever enough to resolve the type variable `a` in the type class otherwise.

src/GraphQL/API/Enum.hs
jml: Well, `unsafeMakeName` panics if it's broken, so we won't have to do error checking either. 

How about having two points of parametrisation? i.e. 

```haskell
class GraphQLEnum a where
   enumValues :: Proxy a -> Either NameError [Name]
   enumValues p = traverse makeName (textEnumValues p)
   textEnumValues :: Proxy a -> [Text]

   enumFromValue :: Name -> Either Text a
   enumToValue :: a -> Name
 ```

Then someone can define:

```haskell
instance GraphQLEnum Foo where
  textEnumValues = ["Bar", "Baz"]
```

And everything is still exactly as fine as it is now.

(NB: I think `textEnumValues` is a bad name).

teh:
Thanks! I'll merge but will keep in mind that we need to fix `enumValues`.

PR #66, opened 2017-01-12
Validation for values and variables, plus substitution

### LINKED ISSUE #58: Resolve variables before delegating to resolver
Currently, the validation code leaves `AST.Value` objects around. These can be variables, which is bad. By the time `Resolver` code gets there, all the variables should be gone.

jml:
This is a pretty big change. I'm sorry that it will make the review harder, but I struggled to separate it into meaningful chunks in my head.

GitHub hides `Value` and `Validation` for me, and I think that might be the best way to review it. Review the stuff that _uses_ those first, then review those two to see how it's all implemented.

The goal was to substitute variables. This meant restructuring `Value` (as discussed), and parametrising most of the types in the `Validation` module by value, to indicate whether they are AST values, valid values that might be variables that don't have definitions, valid values that might be variables that _do_ have definitions, or just concrete values.

One nice thing that falls out of the approach Validation was already taking is that we get the dynamic scoping of GraphQL rather naturally.

The "user-facing" API has been polished a little bit, and you can maybe see hints of this coming together in the examples. We're now exposing the fact that, say, `QueryDocument` is a parametrised type to end users. I'd be OK with adding a bucketload of `newtype` definitions to hide that, but would rather not abandon the type parameter approach in the implementation unless we can find something that provides the same guarantees about the state of our values (Validation is getting rather large, and I need the compiler to help me keep things straight!) and doesn't involve a tonne of duplication.

Fixes #58

### DIFF
--- a/src/GraphQL/Internal/Execution.hs
+++ b/src/GraphQL/Internal/Execution.hs
@@ -103,7 +103,11 @@ formatError (MissingValue name) = "Missing value for " <> show name <> " and mus
 formatError (NoSuchOperation name) = "Requested operation " <> show name <> " but couldn't find it."
 formatError NoAnonymousOperation = "No name supplied for opertaion, but no anonymous operation."
 
--- | Variable values.
+-- | A map of variables to their values.
+--
+-- In GraphQL the variable values are not part of the query itself, they are
+-- instead passed in through a separate channel. Create a 'VariableValues'
+-- from this other channel and pass it to 'substituteVariables'.
 --
 -- GraphQL allows the values of variables to be specified, but doesn't provide
 -- a way for doing so in the language.
--- a/src/GraphQL/Internal/Execution.hs
+++ b/src/GraphQL/Internal/Execution.hs
@@ -1,0 +1,110 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE PatternSynonyms #-}
+-- | Implement the \"Execution\" part of the GraphQL spec.
+--
+-- Actually, most of the execution work takes place in 'GraphQL.Resolver', but
+-- there's still a fair bit required to glue together the results of
+-- 'GraphQL.Internal.Validation' and the processing in 'GraphQL.Resolver'.
+-- This module provides that glue.
+module GraphQL.Internal.Execution
+  ( VariableValues
+  , ExecutionError(..)
+  , formatError
+  , getOperation
+  , substituteVariables
+  ) where
+
+import Protolude hiding (Type)
+
+import qualified Data.Map as Map
+import GraphQL.Value
+  ( Name
+  , Value
+  , pattern ValueNull
+  , Value'(..)
+  , List'(..)
+  , Object'(..)
+  )
+import GraphQL.Internal.Validation
+  ( Operation
+  , QueryDocument(..)
+  , VariableDefinition(..)
+  , VariableValue
+  , Variable
+  , Type(..)
+  )
+
+-- Maybe we should have?
+--
+-- >>> data Request = Request Schema Document (Maybe Operation) (Maybe VariableValues) InitialValue
+--
+-- And then we can have:
+--
+-- execute :: Request -> m Response
+--
+-- Where 'Reponse' is from 'GraphQL.Internal.Output'.
+
+-- | Get an operation from a GraphQL document
+--
+-- <https://facebook.github.io/graphql/#sec-Executing-Requests>
+--
+-- GetOperation(document, operationName):
+--
+--   * If {operationName} is {null}:
+--     * If {document} contains exactly one operation.
+--       * Return the Operation contained in the {document}.
+--     * Otherwise produce a query error requiring {operationName}.
+--   * Otherwise:
+--     * Let {operation} be the Operation named {operationName} in {document}.
+--     * If {operation} was not found, produce a query error.
+--     * Return {operation}.
+getOperation :: QueryDocument value -> Maybe Name -> Either ExecutionError (Operation value)
+getOperation (LoneAnonymousOperation op) Nothing = pure op
+getOperation (MultipleOperations ops) (Just name) = note (NoSuchOperation name) (Map.lookup name ops)
+getOperation (MultipleOperations ops) Nothing =
+  case toList ops of
+    [op] -> pure op
+    _ -> throwError NoAnonymousOperation
+getOperation _ (Just name) = throwError (NoSuchOperation name)
+
+
+-- | Substitute variables in a GraphQL document.
+--
+-- Once this is done, there will be no variables in the document whatsoever.
+substituteVariables :: Operation VariableValue -> VariableValues -> Either ExecutionError (Operation Value)
+substituteVariables op vars = traverse (replaceVariable vars) op
+
+replaceVariable :: VariableValues -> VariableValue -> Either ExecutionError Value
+replaceVariable vars value =
+  case value of
+    ValueScalar' (Left defn) -> getValue defn
+    ValueScalar' (Right v) -> pure (ValueScalar' v)
+    ValueList' (List' xs) -> ValueList' . List' <$> traverse (replaceVariable vars) xs
+    ValueObject' (Object' xs) -> ValueObject' . Object' <$> traverse (replaceVariable vars) xs
+  where
+
+    getValue :: VariableDefinition -> Either ExecutionError Value
+    getValue (VariableDefinition variableName variableType defaultValue) =
+      note (MissingValue variableName) $
+      Map.lookup variableName vars <|> defaultValue <|> allowNull variableType
+
+    allowNull (TypeNonNull _) = empty
+    allowNull _ = pure ValueNull
+
+
+data ExecutionError
+  = MissingValue Variable
+  | NoSuchOperation Name
+  | NoAnonymousOperation
+  deriving (Eq, Show)
+
+formatError :: ExecutionError -> Text
+formatError (MissingValue name) = "Missing value for " <> show name <> " and must be non-null."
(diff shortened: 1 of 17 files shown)


src/GraphQL/Internal/Validation.hs
```diff
@@ -65,72 +71,71 @@ import qualified Data.Set as Set
...
-data Operation
-  = Query [AST.VariableDefinition] Directives SelectionSet
-  | Mutation  [AST.VariableDefinition] Directives SelectionSet
+data Operation value
+  = Query VariableDefinitions (Directives value) (SelectionSet value)
+  | Mutation VariableDefinitions (Directives value) (SelectionSet value)
   deriving (Eq, Show)
 
+instance Functor Operation where
+  fmap f (Query vars directives selectionSet) = Query vars (fmap f directives) (map (fmap f) selectionSet)
```
teh: I assume these three classes can't be derived using `DeriveTraversable` & friends?

src/GraphQL/Internal/Validation.hs
```diff
@@ -386,12 +472,77 @@ resolveFragmentDefinitions allFragments =
           modify (Set.insert name)
           FragmentSpread name directives <$> resolveFragment' definition
 
+-- * Variables
+
+data VariableDefinition = VariableDefinition Variable AST.Type (Maybe Value) deriving (Eq, Ord, Show)
```
teh: Does `Maybe Value` mean the type is nullable? BTW it might be worth having our own `Nullable` instead of `Maybe` because of all the problems that `Maybe` has caused in aeson? The name would also better communicate what we use it for.

src/GraphQL/Internal/Validation.hs
```diff
@@ -386,12 +472,77 @@ resolveFragmentDefinitions allFragments =
...
+validateVariableDefinition (AST.VariableDefinition name varType value) =
+  VariableDefinition name varType <$> (traverse validateDefaultValue value)
+
+-- | Ensure that a default value contains no variables.
+validateDefaultValue :: AST.DefaultValue -> Validation Value
+validateDefaultValue defaultValue =
+  case astToVariableValue defaultValue of
+    Nothing -> throwE $ InvalidValue defaultValue
+    Just value ->
+      for value $ \scalar -> case scalar of
```
teh: Is this a normal use for `for`? Reads a bit weird to traverse a single-element container...

`\scalar -> .. case scalar of ` might be a case for `LambdaCase` if that's your thing.

src/GraphQL/Value.hs
```diff
@@ -33,69 +49,167 @@ module GraphQL.Value
...
+  | ValueList' (List' scalar)
+  | ValueObject' (Object' scalar)
+  deriving (Eq, Ord, Show, Functor)
+
+instance Foldable Value' where
+  foldMap f (ValueScalar' scalar) = f scalar
+  foldMap f (ValueList' values) = foldMap f values
+  foldMap f (ValueObject' obj) = foldMap f obj
+
+instance Traversable Value' where
```
teh: can this be derived automatically?

src/GraphQL/Internal/Execution.hs
```diff
@@ -0,0 +1,110 @@
...
+substituteVariables :: Operation VariableValue -> VariableValues -> Either ExecutionError (Operation Value)
+substituteVariables op vars = traverse (replaceVariable vars) op
+
+replaceVariable :: VariableValues -> VariableValue -> Either ExecutionError Value
+replaceVariable vars value =
+  case value of
+    ValueScalar' (Left defn) -> getValue defn
+    ValueScalar' (Right v) -> pure (ValueScalar' v)
+    ValueList' (List' xs) -> ValueList' . List' <$> traverse (replaceVariable vars) xs
+    ValueObject' (Object' xs) -> ValueObject' . Object' <$> traverse (replaceVariable vars) xs
```
jml: Thanks!

src/GraphQL/Internal/Execution.hs
```diff
@@ -0,0 +1,110 @@
...
+  deriving (Eq, Show)
+
+formatError :: ExecutionError -> Text
+formatError (MissingValue name) = "Missing value for " <> show name <> " and must be non-null."
+formatError (NoSuchOperation name) = "Requested operation " <> show name <> " but couldn't find it."
+formatError NoAnonymousOperation = "No name supplied for opertaion, but no anonymous operation."
+
+-- | Variable values.
+--
+-- GraphQL allows the values of variables to be specified, but doesn't provide
```
jml: Updated. PTAL.

src/GraphQL/Internal/Validation.hs
jml: You know, they might well be. I wrote these because it helped me figure out what I was doing. Because I don't understand how `DeriveTraversable` and friends work, I left them as is.

However, I should learn! I'll write some tests (maybe won't commit them) and do some reading and update.  TODO

src/GraphQL/Internal/Validation.hs
jml: No, the `Value` is a default value, so `Maybe Value` means you don't have to supply a default value. The AST module has a type alias from `Value` to `DefaultValue`—we can do that here if you think it would help.

I agree it's a bit of a shambles in Aeson. To avoid that, we need to be careful with defining `Maybe` instances for `FromValue` or `ToValue` 

Whether or not the value is nullable is encoded in `AST.Type`. You can see an example of how the query document defines the type here: https://facebook.github.io/graphql/#sec-Language.Variables and an example of defaults at https://facebook.github.io/graphql/#sec-Variable-Default-Values-Are-Correctly-Typed.

FWIW, the semantics of "no default value supplied" are that if the user doesn't provide a value and the value is nullable, use `nil`, if it's _not_ nullable, return an error.

This PR implements those semantics in `substituteVariables` but makes no other use of `AST.Type`.

src/GraphQL/Internal/Validation.hs
jml: `for` is `flip traverse`—in this model, a `Value'` can be traversed over its scalars. It's pretty normal to use it when there's a syntactic benefit, e.g. using a lambda expression.

`LambdaCase` is neat, hadn't really tried it before! Switching to that.

src/GraphQL/Value.hs
jml: Don't know. Will do some experiments & update accordingly, as per other comment.

src/GraphQL/Value.hs
```diff
@@ -33,69 +49,167 @@ module GraphQL.Value
...
+  | otherwise = oneof [ ValueScalar' <$> arbitrary
+                      , ValueObject' <$> genObject (n - 1)
+                      , ValueList' . List' <$> listOf (genValue (n - 1))
                       ]
 
+-- | A GraphQL value which contains no variables.
+type Value = Value' ConstScalar
+
+-- TODO: These next two definitions are quite internal. We should move this
+-- mode to Internal and then re-export the bits that end-users will use.
```
jml: Yes, fixed.

src/GraphQL/Value.hs
```diff
@@ -33,69 +49,167 @@ module GraphQL.Value
...
+instance ToJSON ConstScalar where
+  toJSON (ConstInt x) = toJSON x
+  toJSON (ConstFloat x) = toJSON x
+  toJSON (ConstBoolean x) = toJSON x
+  toJSON (ConstString x) = toJSON x
+  toJSON (ConstEnum x) = toJSON x
+  toJSON ConstNull = Aeson.Null
+
+-- | A value which contains no other values, and might be a variable that
+-- might lack a definition.
```
jml: Spec reference is here: https://facebook.github.io/graphql/#VariableDefinition. I've updated the comment. I hope it makes it more clear. 

I thought about "binding" or using "bound" in the comment, but I'm unsure.

A variable is this:

```graphql
$variable
```

A variable definition is this:

```graphql
query foo($variable: Int) {
  ...
}
```

And a value can be provided for a variable with something like:

```json
{
  "variable": 42
}
```

At which stage is the binding?

src/GraphQL/Internal/Validation.hs
teh: Ah I see now, thanks. Can you copy-paste your first paragraph as a comment? Maybe it's worth making this a record so we can name the fields even if we don't use the accessors?

src/GraphQL/Internal/Execution.hs
```diff
@@ -0,0 +1,110 @@
...
+  , Type(..)
+  )
+
+-- Maybe we should have?
+--
+-- >>> data Request = Request Schema Document (Maybe Operation) (Maybe VariableValues) InitialValue
+--
+-- And then we can have:
+--
+-- execute :: Request -> m Response
```
teh: That could work though the use would be a bit weird:

```haskell
let executor = execute @MiniAPI miniHandler
for incoming $ \request -> do
   executor request
```

That's why it's currently called `buildResover` so we can do `let resolve = buildResolver @MiniAPI miniHandler`.

But also not sure. We can try a few things to see what works.

src/GraphQL/Value.hs
teh: Right, probably fine as it is. Thanks for explaining.

src/GraphQL/Internal/Execution.hs
jml: You don't need the `do`. That code is equivalent to:

```haskell
for incoming $ execute @MiniAPI miniHandler
```

src/GraphQL/Internal/Validation.hs
jml: Done. I'm fond of record syntax in general, both for the extra documentation, and because there are times when accessors are really handy to have.

jml:
Merging without removing `Foldable` et al. Will try that soon.

PR #69, opened 2017-01-13
Generic from value

teh:
wait until your merge before reviewing

### DIFF
--- a/src/GraphQL/API.hs
+++ b/src/GraphQL/API.hs
@@ -238,7 +238,7 @@ instance forall t. (HasAnnotatedType t) => HasAnnotatedType (List t) where
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedType (Enum ks enum) where
   getAnnotatedType = do
     let name = nameFromSymbol @ks
-    let enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
+    let enums = sequenceA (enumValues @enum) :: Either NameError [Name]
     let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedType . TypeDefinitionEnum <$> et
 
@@ -291,7 +291,7 @@ instance forall t. (HasAnnotatedInputType t) => HasAnnotatedInputType (List t) w
 instance forall ks enum. (KnownSymbol ks, GraphQLEnum enum) => HasAnnotatedInputType (Enum ks enum) where
   getAnnotatedInputType = do
     let name = nameFromSymbol @ks
-        enums = sequenceA (enumValues @enum Proxy) :: Either NameError [Name]
+        enums = sequenceA (enumValues @enum) :: Either NameError [Name]
     let et = EnumTypeDefinition <$> name <*> map (map EnumValueDefinition) enums
     TypeNonNull . NonNullTypeNamed . DefinedInputType . InputTypeDefinitionEnum <$> et
 
--- a/src/GraphQL/API.hs
+++ b/src/GraphQL/API.hs
@@ -1,7 +1,13 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE FlexibleContexts #-}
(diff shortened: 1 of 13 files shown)


src/GraphQL/API.hs
```diff
@@ -26,14 +32,15 @@ module GraphQL.API
   , getAnnotatedInputType
   ) where
 
-import Protolude hiding (Enum)
+import Protolude hiding (Enum, TypeError)
```
jml: At a guess, the `TypeError` you're hiding is the same as the one in `TypeLits`.

src/GraphQL/API.hs
```diff
@@ -282,6 +291,53 @@ instance forall t. (HasAnnotatedInputType t) => HasAnnotatedInputType (List t) w
...
+
+instance forall dataName consName records s l p.
+  ( KnownSymbol dataName
+  , KnownSymbol consName
+  , GenericInputObjectFieldDefinitions records
+  ) => GenericAnnotatedInputType (D1 ('MetaData dataName s l 'False)
+                                  (C1 ('MetaCons consName p 'True) records
+                                  )) where
+  genericGetAnnotatedInputType = do
+    name <- nameFromSymbol @dataName
```
jml: For our _generic_ instances, I think it might be OK to assume that the names are valid GraphQL names.

Providing that:
- [ ] the symbol names come from Haskell function / variable names
- [ ] Haskell function / variable names are valid GraphQL names
- [ ] we document this, probably by having a dedicated `haskellNameToGraphQLName :: KnownSymbol ks -> Name` with a comment saying that you should only use this for known haskell names
- [ ] the non-generic type class still allows `NameError` failures (because we have no idea where they are getting their names from) 

What do you think?

src/GraphQL/API.hs
```diff
@@ -282,6 +291,53 @@ instance forall t. (HasAnnotatedInputType t) => HasAnnotatedInputType (List t) w
...
+    pure (l:r)
+
+instance forall wrappedType fieldName u s l.
+  ( KnownSymbol fieldName
+  , HasAnnotatedInputType wrappedType
+  ) => GenericInputObjectFieldDefinitions (S1 ('MetaSel ('Just fieldName) u s l) (Rec0 wrappedType)) where
+  genericGetInputObjectFieldDefinitions = do
+    name <- nameFromSymbol @fieldName
+    annotatedInputType <- getAnnotatedInputType @wrappedType
+    let l = InputObjectFieldDefinition name annotatedInputType Nothing
```
jml: I had completely forgotten about this kind of defaulting! (No action required).

jml:
Gah, sorry, the conflicts are my fault (again!)

src/GraphQL/API.hs
jml: https://facebook.github.io/graphql/#sec-Input-Objects documents their behaviour, fwiw.

src/GraphQL/Resolver.hs
```diff
@@ -41,6 +41,8 @@ import GraphQL.API
   )
 import qualified GraphQL.API as API
 import qualified GraphQL.Value as GValue
+import GraphQL.Value.FromValue (FromValue(..))
+import GraphQL.Value.ToValue (ToValue(..))
 import GraphQL.Value (Value)
```
teh: I don't think so, because that introduces a circular dependency (`FromValue` imports `Value`).

src/GraphQL/API.hs
teh: Not sure what the `Protolude` one is for, but it's not the same.

src/GraphQL/API.hs
teh: We could do that but I'm not sure what it buys us: We need to return an error anyway for `HasAnnotatedType`. I.e. what do you suggest we do instead of the current implementation?

teh:
Sent back with a few questions & resolved conflicts.

teh:
merging now though I am fully aware that the API isn't shippable as-is. What to re-export in top-level GraphQL also an interesting question TBD.

src/GraphQL/API.hs
jml: I thought I suggested it, but I'll have a stab separately. I think it buys us a slightly cleaner internal generic implementation, and doesn't buy our users very much.

PR #72, opened 2017-01-14
Format error typeclass

jml:
As mentioned in #66. 

Lots of human-readable errors.

I haven't actually verified any of them, but I reckon this is an improvement, and should signal clearly to our users that we care about this sort of thing.

Changed the union handling a bit because I couldn't understand the error conditions well enough. See commit message on c0e0f67 for rationale.

### DIFF
--- a/src/GraphQL/API/Enum.hs
+++ b/src/GraphQL/API/Enum.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DefaultSignatures #-}
-{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE KindSignatures #-}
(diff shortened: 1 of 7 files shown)


src/GraphQL/API/Enum.hs
```diff
@@ -50,6 +56,14 @@ instance forall conName f p b.
   genericEnumToValue (L1 _) =
     case nameFromSymbol @conName of
       Right name -> name
+      -- XXX: This is impossible to catch during validation, because we cannot
```
teh: I don't think this statement is true. When we run `genericEnumValues` we do the same name-check so all we need to do is make sure `genericEnumValues` runs for validation, which it does I think if we `deepseq` the value representation.

src/GraphQL/Internal/Validation.hs
```diff
@@ -625,6 +626,23 @@ data ValidationError
   | InvalidDefaultValue AST.Value
   deriving (Eq, Show)
 
+instance GraphQLError ValidationError where
+  formatError (DuplicateOperation name) = "More than one operation named '" <> show name <> "'"
+  formatError (MixedAnonymousOperations n names)
+    | n > 1 && null names = "Multiple anonymous operations defined. Found " <> show n
+    | otherwise = "Document contains both anonymous operations (" <> show n <> ") and named operations (" <> show names <> ")"
+  formatError (DuplicateArgument name) = "More than one argument named '" <> show name <> "'"
+  formatError (DuplicateFragmentDefinition name) = "More than one fragment named '" <> show name <> "'"
```
teh: One of the lessons learned from the psql client libraries was that it's useful to have a unique name somewhere in the `Error`, because if you don't people start matching strings to detect the error type, making the string representation your API. I'm not sure how we can fit this into `Error` but I think I'd like to keep that in mind.

jml: Interesting point. I'll keep it in mind.

src/GraphQL/API/Enum.hs
jml: That still relies on someone remember to run that code first before calling this code, right?

PR #74, opened 2017-01-16
Make list handlers themselves monadic on top of the elements.

teh:
I've been going back and forth on this one for a while. There are good arguments for and against, but I'm 60:40 now on making the list handler itself monadic so here's the branch.

### DIFF
--- a/src/GraphQL/Resolver.hs
+++ b/src/GraphQL/Resolver.hs
@@ -193,11 +193,13 @@ instance forall m. (Functor m) => HasGraph m Text where
   buildResolver handler _ =  map (ok . toValue) handler
 
 
-instance forall m hg. (Applicative m, HasGraph m hg) => HasGraph m (API.List hg) where
-  type Handler m (API.List hg) = [Handler m hg]
-  buildResolver handler selectionSet = map aggregateResults a
-    where
-      a = traverse (flip (buildResolver @m @hg) selectionSet) handler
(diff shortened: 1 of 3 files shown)


src/GraphQL/Resolver.hs
```diff
@@ -193,11 +193,13 @@ instance forall m. (Functor m) => HasGraph m Text where
   buildResolver handler _ =  map (ok . toValue) handler
 
 
-instance forall m hg. (Applicative m, HasGraph m hg) => HasGraph m (API.List hg) where
-  type Handler m (API.List hg) = [Handler m hg]
-  buildResolver handler selectionSet = map aggregateResults a
-    where
-      a = traverse (flip (buildResolver @m @hg) selectionSet) handler
+instance forall m hg. (Monad m, Applicative m, HasGraph m hg) => HasGraph m (API.List hg) where
+  type Handler m (API.List hg) = m [Handler m hg]
```
jml: Unsure about this. Could be a good idea, could be otherwise.

If this were "regular" Haskell code, and we didn't have the pull of wanting a composable API, we'd probably do something like:

`m [Handler hg]`

Because, for example, if you want to have some code return an `IO [a]`, even if each individual `a` required some monadic action, you could still build the `IO [a]` by sequencing.

However, this _isn't_ regular Haskell code, and we _do_ want composability.  So, I don't know.

jml: We talked in person. We can't eliminate the inner one, and we need the outer one for, e.g., database queries.

PR #76, opened 2017-01-16
Top level API

### LINKED ISSUE #60: Top-level function to take queries and return responses
We have parsing, validation, resolving, and a response object. We need to string them all together.

Notes:

the 'schema' is defined by the server types, an initial value, and any supported directives (although we'll probably skip those for first release) -- so that's the server's input

the client's input is a query document, a variable map, and an optional name of an operation to run. Not sure how these are actually delivered to us.

the output is the Response object in our Output.hs module

overall process is:
  1. parse query (if it fails, respond with PreExecutionFailure)
  2. transform the AST into something valid, ignoring type checks, but making sure that arguments aren't duplicated, no circular references. (if fail, PreExecutionError)
  Note that we don't need the schema *or* the variables yet.
  3. substitute in the variables. now everything is literal.
  4. apply the directives. probably hard-code @skip & @include for the first release. if fail, execution error, and continue with fragments without bad directives
  We now have something that's just objects & fragments
  5. "collect the fields" to avoid duplicate processing (c.f. https://facebook.github.io/graphql/#sec-Field-Collection). still not 100% sure about this.
  6. delegate to server-supplied code, collecting any results and errors
  7. wrap whatever we've got off in a Response

jml:
As discussed, new functions in top-level `GraphQL` module:

 * `executeQuery`
 * `interpretQuery`
 * `interpretAnonymousQuery`

These all have copious (but not necessarily good) documentation. Please check the Haddock documentation generated by Circle to confirm that it's readable in context.

I've deleted the filesystem example, as discussed. I've updated the other examples to refer to `interpretAnonymousQuery`, along with making some other changes to improve their documentation.

`getOperation` and `SelectionSet` are no longer exposed at the top-level.

`TypeApiTests` (which should really be renamed to `ResolverTests`) also got a bit of a tidy up. I'm not sure what the point of the custom monad is any more, but it's there anyway.

Fixes #60


src/GraphQL.hs
```diff
@@ -54,6 +69,61 @@ instance GraphQLError QueryError where
...
+  formatError (NonObjectResult v) =
+    "Query returned a value that is not an object: " <> show v
+
+-- | Execute a GraphQL query.
+executeQuery
+  :: forall api m. (HasGraph m api, Applicative m)
+  => Handler m api -- ^ Handler for the query. This links the query to the code you've written to handle it.
+  -> QueryDocument VariableValue  -- ^ A validated query document. Build one with 'compileQuery'.
+  -> Maybe Name -- ^ An optional name. If 'Nothing', then executes the only operation in the query. If @Just "something"@, executes the query named @"something".
+  -> VariableValues -- ^ Values for variables defined in the query document. A map of 'Variable' to 'Value'.
```
teh: Maybe add how to construct `VariableValues` in the docstring?

tests/Examples/InputObject.hs
```diff
@@ -26,26 +25,21 @@ root :: Handler IO Query
 root = pure (\dogStuff -> pure (show dogStuff))
 
 -- $setup
--- >>> import qualified GraphQL.Internal.Encoder as Encode
--- >>> import GraphQL.Value (valueToAST)
+-- >>> import Data.Aeson (encode)
+-- >>> import GraphQL.Value.ToValue (ToValue(..))
+
+-- TODO: jml thinks it's a bit confusing to have `show` output in these
```
teh: not sure what the TODO is here. Tests seem fine to me?

src/GraphQL.hs
```diff
@@ -54,6 +69,61 @@ instance GraphQLError QueryError where
     "Validation errors:\n" <> mconcat ["  " <> formatError e <> "\n" | e <- NonEmpty.toList es]
   formatError (ExecutionError e) =
     "Execution error: " <> show e
+  formatError (NonObjectResult v) =
+    "Query returned a value that is not an object: " <> show v
+
+-- | Execute a GraphQL query.
+executeQuery
+  :: forall api m. (HasGraph m api, Applicative m)
```
teh: We discussed adding an `Object n i f ~ api` constraint to make sure `executeQuery` can run only on Objects. Up to you.

src/GraphQL.hs
jml: I started to do this and realised that we don't have a great story for it. Separate PR.

tests/Examples/InputObject.hs
jml: The braces & quotes within the string within the JSON confuse me a little syntactically. Hard to explain beyond what I've already commented. Have just done the TODO in 88056e4. Let me know if a) that clarifies what I meant by the TODO, b) you think that makes the example more effective.

src/GraphQL.hs
jml: Separate PR? Also, overnight I thought that that an `IsQueryRoot` constraint (with a `resolveRoot :: Handler m a -> m Object` method) wouldn't be such a bad thing.

tests/Examples/InputObject.hs
jml: Will merge this change as-is, but please do let me know if you want it changed again.

src/GraphQL.hs
alunduil: I'm looking for how to accomplish this now, but don't see any new information.  Is the mentioned PR available?

PR #77, opened 2017-01-17
End-to-end tests

jml:
As discussed, end-to-end tests. By no means a full battery, but it should be enough to give the general idea.

I rebased the previous PR and incorporated @teh's fixes from #78. 

The "nested queries" test asserts a bad result. Asserting the correct result would fail. I figure this is the best way to document the behaviour. I'm pretty sure it's a release blocker, but I'd like to get this PR in more-or-less as is, without any further scope expansion.

### DIFF
--- a/src/GraphQL/Resolver.hs
+++ b/src/GraphQL/Resolver.hs
@@ -193,6 +193,10 @@ instance forall m. (Functor m) => HasGraph m Text where
   -- TODO check that selectionset is empty (we expect a terminal node)
   buildResolver handler _ =  map (ok . toValue) handler
 
+instance forall m. (Functor m) => HasGraph m Bool where
+  type Handler m Bool = m Bool
+  -- TODO check that selectionset is empty (we expect a terminal node)
+  buildResolver handler _ =  map (ok . toValue) handler
 
 instance forall m hg. (Monad m, Applicative m, HasGraph m hg) => HasGraph m (API.List hg) where
   type Handler m (API.List hg) = m [Handler m hg]
--- a/src/GraphQL/Resolver.hs
+++ b/src/GraphQL/Resolver.hs
@@ -43,7 +43,11 @@ import GraphQL.API
   )
 import qualified GraphQL.API as API
 import qualified GraphQL.Value as GValue
-import GraphQL.Value (Name, Value)
+import GraphQL.Value
+  ( Name
+  , Value
+  , pattern ValueEnum
+  )
 import GraphQL.Value.FromValue (FromValue(..))
 import GraphQL.Value.ToValue (ToValue(..))
 import qualified GraphQL.Internal.AST as AST
@@ -309,9 +313,9 @@ instance forall ksK t f m name.
   buildFieldResolver handler field = do
     argName <- first SchemaError (AST.nameFromSymbol @ksK)
     value <- case lookupArgument field argName of
-      Nothing -> panic "No"
-      -- TODO: extract enum name and use instead of argName!
-      Just _ -> first (InvalidValue argName) (API.enumFromValue @t argName)
+      Nothing -> valueMissing @t argName
+      Just (ValueEnum enum) -> first (InvalidValue argName) (API.enumFromValue @t enum)
+      Just value -> Left (InvalidValue argName (show value <> " not an enum: " <> show (API.enumValues @t)))
     buildFieldResolver @m @f (handler value) field
 
 -- Note that we enumerate all ks variables with capital letters so we
--- a/src/GraphQL/Resolver.hs
+++ b/src/GraphQL/Resolver.hs
@@ -344,9 +344,13 @@ type family RunFieldsHandler (m :: Type -> Type) (a :: Type) = (r :: Type) where
 
 
 class RunFields m a where
-  -- runFields runs a a single AST.Selection over all possible fields
-  -- (as specified by the type), returing exactly one ObjectField when
-  -- a field matches, or an error otherwise.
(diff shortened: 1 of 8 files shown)


tests/EndToEndTests.hs
```diff
@@ -0,0 +1,113 @@
...
+  -- /Users/jml/src/graphql-api/tests/EndToEndTests.hs:59:3: error:
+  --     • The type family ‘Handler’ should have 2 arguments, but has been given 1
+  --     • In the first argument of ‘(.)’, namely ‘pure @(Handler IO)’
+  --       In the first argument of ‘(:<>)’, namely
+  --         ‘pure @(Handler IO) . doesKnowCommand dog’
+  --       In the second argument of ‘(:<>)’, namely
+  --         ‘pure @(Handler IO) . doesKnowCommand dog
+  --          :<>
+  --            pure @(Handler IO) . isHouseTrained dog :<> viewServerHuman owner’
+  pure @(Handler IO) . doesKnowCommand dog :<>
```
teh: the main issues apart from enum not working ATM are

1. The type signature isn't `m (ServerDog -> DogCommand -> Bool)` it is `ServerDog -> DogCommand -> m Bool`.
2. `Int` doesn't have a FromValue ATM, use `Int32`

The `pure` calls should not require explicit type applications (`@`) once the handler type matches.

src/GraphQL/Value/ToValue.hs
```diff
@@ -21,6 +21,12 @@ instance ToValue (Value' ConstScalar) where
 instance ToValue a => ToValue [a] where
   toValue = toValue . List' . map toValue
 
+-- TODO - tom still thinks that using Maybe for nullable is maybe not
+-- the best idea.
```
jml: I'm starting to agree with you on this. Also think that having `Maybe a` be `Defaultable` is not a great idea.

Reasoning: the GraphQL spec says value look-up goes:
- provided value, if not there then
- default value, if not there then
- null, if nullable, otherwise
- error

Nullability should be held as a strongly distinct concept from defaultability.

src/GraphQL/Resolver.hs
```diff
@@ -241,92 +254,133 @@ resolveField handler nextHandler field =
...
+  RunFieldsHandler m (a :> b) = FieldHandler m (FieldResolverDispatchType (a :> b))
   RunFieldsHandler m a = TypeError (
     'Text "Unexpected RunFieldsHandler types: " ':<>: 'ShowType a)
 
 
 class RunFields m a where
-  -- runFields is run on a single AST.Selection so it can only ever
-  -- return one ObjectField.
+  -- runFields runs a a single AST.Selection over all possible fields
+  -- (as specified by the type), returing exactly one ObjectField when
```
teh: "returning" - I can't spell :P

tests/ExampleSchema.hs
```diff
@@ -0,0 +1,304 @@
...
+--
+-- Unlike the spec, we don't define a @QueryRoot@ type here, instead
+-- encouraging test modules to define their own as appropriate to their needs.
+--
+-- We'll repeat bits of the schema below, explaining how they translate into
+-- Haskell as we go.
+
+module ExampleSchema
+  ( DogCommand
+  , DogCommandEnum
```
teh: don't think we still need to export `DogCommand`

tests/ExampleSchema.hs
```diff
@@ -0,0 +1,304 @@
...
+-- To define this in Haskell we need to do three things:
+--
+--  1. Define a sum type with nullary constructors to represent the enum
+--     (here, 'DogCommandEnum')
+--  2. Make it an instance of 'GraphQLEnum'
+--  3. Wrap the sum type in 'Enum', e.g. @Enum "DogCommand" DogCommandEnum@
+--     so it can be placed in a schema.
+data DogCommandEnum = Sit | Down | Heel deriving (Show, Eq, Ord, Generic)
+
+instance GraphQLEnum DogCommandEnum
```
teh: I'd probably put the `Enum` suffix onto the actual `Enum`

tests/ExampleSchema.hs
```diff
@@ -0,0 +1,304 @@
...
+-- | A dog.
+--
+-- This is an example of a GraphQL \"object\".
+--
+-- @
+-- type Dog implements Pet {
+--   name: String!
+--   nickname: String
+--   barkVolume: Int
+--   doesKnowCommand(dogCommand: DogCommand!): Boolean!
```
teh: Is `dogCommand` meant to be nullable?

src/GraphQL/Resolver.hs
jml: Fixed. Also, made this a doc comment (`-- |` rather than `--`) and made some markup & wording tweaks.

tests/ExampleSchema.hs
jml: Agreed. Have updated the SchemaTests to just construct `Enum "DogCommand" DogCommand` themselves.

tests/ExampleSchema.hs
jml: Good call. Done. I've also removed the type alias completely, instead embedding the `Enum "DogCommand" DogCommand"` in the schema. Because it's only there once, it works, and reduces the confusion about which one I mean.

tests/ExampleSchema.hs
jml: No, it isn't. In the example schema, `dogCommand` is the non-nullable argument example (hence the `!`), and `isHouseTrained` is the nullable example.

Issue #59, opened 2017-01-02
Enforce field uniqueness in selections


jml:

teh:
I'm not super sure what this means. Is this to catch a query like this:

```
{
  a
  a
  b
}
```

?

jml:
It's kind of detailed actually.  I'll try to sum up:

## Handle fragment spreads

e.g. This is a valid query:

```graphql
{
  foo
  bar
  ... myFrag
}

fragment myFrag {
  bar
  qux
}
```

And the output should look something like:

```json
{
  "foo": 1,
  "bar": 2,
  "qux": 3,
}
```

## Aggregate identical fields

e.g. 

```graphql
{
  a
  b
  b
}
```

Should just be processed as if it were:

```graphql
{
  a
  b
}
```

## Bork on non-identical fields

```graphql
{
  a
  b
  b(1)
}
```

jml:
Relevant parts of the spec:

- https://facebook.github.io/graphql/#sec-Field-Selection-Merging
- https://facebook.github.io/graphql/#sec-Field-Collection
- https://facebook.github.io/graphql/#sec-Value-Completion (esp. `MergeSelectionSet`)

Reason all these are together is because the implementation for all of them will involve changing the type of `SelectionSet` in validation. Currently, it's a list of `Selection`, but it'll probably become some sort of `OrderedMap` or at least something more structured. 

Probably also `FragmentSpread` will disappear from the fully-resolved type, which will be a mercy.

Anyway, as you can guess from this commentary, I'm at the level where it's easier for me to code it than to explain it.

jml:
Also!

```graphql
{
  a {
    foo
  }
  a {
    bar
  }
}
```

=> 

```json
{
  "a": {
    "foo": 1,
    "bar": 2
  }
}
```

jml:
The way this works in the spec is that _validation_ confirms that everything is mergeable and at execution each selection set gets flattened to fields only (one level at a time), and then evaluated.

The natural thing to do in our approach would be to flatten everything in the validator, but we this doesn't chime with the way we're doing unions (which I understand less now than I did before I started reading the spec).

Will continue to mull.

jml:
Here's where I've got to conceptually. Don't know how this fits into our code.

- after validation, each selection set will be a thing that contains:
  - fields
  - inline fragments with type conditions
  - no fragment spreads
  - i.e. a slightly narrower `Selection` (with no need for `spread` type parameter!)
[...]

PR #84, opened 2017-01-20
Make type condition optional for inline fragments

### LINKED ISSUE #70: Type conditions on inline fragment should be optional
See https://facebook.github.io/graphql/#sec-Inline-Fragments

It's valid to have an inline fragment with no type condition (for use with directives, say).

We currently treat this as mandatory, even at the AST level.

jml:
Fixes #70. See https://facebook.github.io/graphql/#sec-Inline-Fragments.

Doing this now because I want to work on #59, and it'll be easier for me if I'm working from a known correct base. 

This PR depends on #83 (and thus #82). Probably review & merge those first before this one.

### DIFF
--- a/graphql-api.cabal
+++ b/graphql-api.cabal
@@ -40,14 +40,15 @@ library
       GraphQL.API
       GraphQL.API.Enum
       GraphQL.Internal.Arbitrary
-      GraphQL.Internal.AST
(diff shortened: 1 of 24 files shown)


jml:
Yeah, me too.

PR #85, opened 2017-01-20
Resolve field name before checking field arguments. Closes #81.

### LINKED ISSUE #81: Cannot distinguish between missing arguments and field not wanted
As per #77 

On the example schema, this query:

```graphql
{
  name
  owner {
    name
  }
}
```

Returns:

```json
{
  "data": {
    "dog": {
      "name": "Mortgage",
      "owner": null
    }
  },
  "errors": [
    {
      "message": "No value provided for Name {getNameText = \"dogCommand\"}, and no default specified."
    }
  ]
}
```

I would expect it to return:

```json
{
  "data": {
    "dog": {
      "name": "Mortgage",
      "owner": {
        "name": "jml"
      }
    }
  }
}
```

teh:
Before this change we tried to read field arguments even if the query
never requested the actual field itself. This is obviously wrong
because we'd never supply arguments for fields we're not requesting.

Even if all fields were optionsal we'd still waste CPU cycles.

### DIFF
--- a/src/GraphQL/Resolver.hs
+++ b/src/GraphQL/Resolver.hs
@@ -67,7 +67,7 @@ data ResolverError
   -- | There was a problem in the schema. Server-side problem.
   = SchemaError NameError
   -- | Couldn't find the requested field in the object. A client-side problem.
-  | FieldNotFoundError (Field Value)
+  | FieldNotFoundError Name
   -- | No value provided for name, and no default specified. Client-side problem.
   | ValueMissing Name
   -- | Could not translate value into Haskell. Probably a client-side problem.
@@ -85,7 +85,7 @@ instance GraphQLError ResolverError where
   formatError (SchemaError e) =
     "Schema error: " <> formatError e
   formatError (FieldNotFoundError field) =
-    "Could not find value for field: " <> show field
+    "Field not supported by the API: " <> show field
   formatError (ValueMissing name) =
     "No value provided for " <> show name <> ", and no default specified."
   formatError (InvalidValue name text) =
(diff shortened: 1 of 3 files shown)


src/GraphQL/Resolver.hs
```diff
@@ -227,63 +227,76 @@ instance forall m ksN enum. (Applicative m, API.GraphQLEnum enum) => HasGraph m
...
-          Result errs value <- resolver
-          pure (Result errs (Just (GValue.ObjectField queryFieldName value)))
+    runResolver :: Name -> m ResolveFieldResult
+    runResolver name'
+      | queryFieldName == name' =
+          case buildFieldResolver @m @dispatchType handler field of
+            Left err -> pure (Result [err] (Just (GValue.ObjectField queryFieldName GValue.ValueNull)))
+            Right resolver -> do
+              Result errs value <- resolver
+              pure (Result errs (Just (GValue.ObjectField queryFieldName value)))
```
jml: Optional. Collapse `runResolver` into the `case`, using guard syntax to check if `queryFieldName` matches `name'`

tests/ResolverTests.hs
```diff
@@ -48,7 +48,7 @@ tests = testSpec "TypeAPI" $ do
       Right (PartialSuccess _ errs) <- runExceptT (interpretAnonymousQuery @T tHandler "{ not_a_field }")
       -- TODO: jml thinks this is a really bad error message. Real problem is
       -- that `not_a_field` was provided.
-      errs `shouldBe` singleError (ValueMissing (unsafeMakeName "x"))
+      errs `shouldBe` singleError (FieldNotFoundError (unsafeMakeName "not_a_field"))
```
jml: This completes the `TODO` comment above, please remove.

jml: This completes the `TODO` comment above, please remove.

jml: This completes the TODO comment above. Please delete it.

teh: done

src/GraphQL/Resolver.hs
teh: I'm not entirely sure what you mean, sorry! Would you mind folding the change into your next branch?

PR #86, opened 2017-01-20
More tests

jml:
Super easy to add these, although I discovered & fixed some bugs while at it.

Also identified some weak spots in the API. Rather than fixing these, which would take brain power I simply don't have right now, I've annotated them with TODOs.

### DIFF
--- a/tests/EndToEndTests.hs
+++ b/tests/EndToEndTests.hs
@@ -8,10 +8,13 @@ module EndToEndTests (tests) where
 
 import Protolude
 
-import Data.Aeson (toJSON, object, (.=))
-import GraphQL (interpretAnonymousQuery)
+import Data.Aeson (Value(Null), toJSON, object, (.=))
+import qualified Data.Map as Map
+import GraphQL (compileQuery, executeQuery, interpretAnonymousQuery, interpretQuery)
 import GraphQL.API (Object, Field)
+import GraphQL.Internal.Syntax.AST (Variable(..))
 import GraphQL.Resolver ((:<>)(..), Handler)
+import GraphQL.Value (makeName)
 import GraphQL.Value.ToValue (ToValue(..))
 import Test.Tasty (TestTree)
 import Test.Tasty.Hspec (testSpec, describe, it, shouldBe)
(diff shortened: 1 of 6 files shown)


tests/EndToEndTests.hs
```diff
@@ -152,3 +155,137 @@ tests = testSpec "End-to-end tests" $ do
...
+              [ "dog" .= object
+                [ "name" .= ("Mortgage" :: Text)
+                ]
+              ]
+            ]
+      toJSON (toValue response) `shouldBe` expected
+    describe "Handles variables" $ do
+      let root = pure (viewServerDog mortgage)
+      let Right query =
+            compileQuery [r|query myQuery($whichCommand: DogCommand) {
```
teh: Is the type name validated? I.e. what happens when you use `DogNoExistCommand`?

jml: No, it's not. That's https://facebook.github.io/graphql/#sec-All-Variable-Usages-are-Allowed, which is listed as being out of scope for the `Validation` module.

I think we could & should bring it back into scope for later releases. 

Tangentially, I think it's interesting to know what typeless GraphQL validation/compilation looks like—I wonder if the GraphQL authors even know.

tests/EndToEndTests.hs
```diff
@@ -152,3 +155,137 @@ tests = testSpec "End-to-end tests" $ do
...
+              ]
+        toJSON (toValue response) `shouldBe` expected
+      it "Substitutes variables when they are provided" $ do
+        -- TODO: This is a crummy way to make a variable map. jml doesn't want
+        -- to come up with a new API in this PR, but probably we should have a
+        -- very simple function to turn a JSON value / object into the
+        -- variable map that we desire. Alternatively, we should have APIs
+        -- like Aeson does.
+        let Right varName = makeName "whichCommand"
+        let vars = Map.singleton (Variable varName) (toValue Sit)
```
teh: This `toValue` is the reason we need a `toValue` instance for `DogCommand` in this specific case.

I am thinking for the general case we might want to use an extra `Enum` in the schema:

```haskell
Object "GoodDoggie" '[] '[Field "favouriteDogCommand" (Enum "DogCommand" DogCommand)]
```

jml: I understand the first sentence (although we'd probably need a `toValue` if something returned an enum).

I don't understand the second sentence & code example at all. Don't we _already_ put the `Enum "Foo"` annotation in the schema?

teh: In ExampleSchema.hs we use `Enum "Foo"` for an argument. For outputs its kinda unspecified ATM, so we're still free to chose the API

jml: Gotcha, thanks for clarifying.

I guess this conversation has made it clear to me that `Enum "Foo"` is pretty much only an annotation for the schema plus a mean of getting all the possible values. I wonder whether our existing typeclass is doing too much: why should it care about conversion at all?

teh: That's a good point. I think the main idea was that we could get out a standard sum-type `data A = A | B | C | ...`  for programming in our handlers.

Without a sum type no pattern matching etc. Haskell doesn't (yet?) allow constructing custom sum types from another type (like `Enum`).

*Iff* we believe that having a closed sum type for handlers is better than having an open set of strings (the only alternative with `Symbol` kinds) then I don't see how we'd get around the `GraphQLEnum` type class. And once we have that class I think it makes little sense to specify possible values independently of the actual sum type (that can lead to drift e.g.).

PR #87, opened 2017-01-21
Rename buildResolver

### LINKED ISSUE #71: Rename 'buildResolver'
Naming something to imply that it'll get used partially sometimes is not very Haskell.

e.g. it's `map`, not `buildMapper`. it's `filter`, not `buildFilterer`.

I suggest `execute` or `resolve`, or something similarly direct.

jml:
Fixes #71 

Also tidied up some TODOs, but otherwise kept scope of branch small.

### DIFF
--- a/src/GraphQL/Resolver.hs
+++ b/src/GraphQL/Resolver.hs
@@ -15,7 +15,7 @@
 
 module GraphQL.Resolver
   ( ResolverError(..) -- XXX: Exporting constructor for tests. Not sure if that's what we really want.
-  , HasGraph(..)
+  , HasResolver(..)
   , (:<>)(..)
   , BuildFieldResolver(..)
   , Defaultable(..)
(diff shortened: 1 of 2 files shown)


src/GraphQL/Resolver.hs
```diff
@@ -217,24 +217,11 @@ instance forall m ksN enum. (Applicative m, API.GraphQLEnum enum) => HasResolver
...
---
--- TODO: Objects. Maybe implement some Generic object reader? I.e. if I do
--- data Greet = Greet { name :: Text, score :: Int } deriving Generic
--- then "instance ReadValue Greet" would fall back on a default reader that
--- expects Objects?
--- Maybe we can use advanced fallbacks like these:
--- https://wiki.haskell.org/GHC/AdvancedOverlap
-
-
--- Iterate through handlers (zipped together with their type
```
teh: this is more of a comment about what we do, but maybe I'll write that up in some other form later

PR #89, opened 2017-01-24
Docs2

teh:


docs/source/tutorial/Introduction.lhs
```diff
@@ -12,63 +12,146 @@ module Introduction where
...
-type HelloWorld = Object "HelloWorld" '[]
-  '[ Argument "greeting" Text :> Field "me" Text
-   ]
-```
+ 1. A schema that defines the service
+ 2. Some code that implements the service's behavior
 
-The example above is equivalent to the following GraphQL type:
+We're going to build a very simple service that says hello to
+people. Our GraphQL schema for this looks like:
```
teh: maybe add a note about not confusing graphql types and haskell types? I got a bit confused with the string even though I wrote part of this :)

docs/source/tutorial/Introduction.lhs
```diff
@@ -12,63 +12,146 @@ module Introduction where
...
+Breaking this down, we define a new Haskell type `Hello`, which is a GraphQL
+object (also named `"Hello"`) that implements no interfaces (hence `'[]`). It
+has one field, called `"greeting"` which returns some `Text` and takes a
+single named argument `"who"`, which is also `Text`.
+
+There are some noteworthy differences between this schema and the GraphQL
+schema:
+
+* The GraphQL schema requires a special annotation to say that a value cannot
+  be null, `!`. In Haskell, we instead assume that everything can't be null.
```
teh: "everything can't be" -> "nothing can be" ?

docs/source/tutorial/Introduction.lhs
```diff
@@ -12,63 +12,146 @@ module Introduction where
...
+
+The first layer of the handler, `pure greeting`, produces the `Hello` object.
+The `pure` might seem redundant here, but making this step monadic allows us
+to run actions in the base monad.
+
+The second layer of the handler, the implementation of `greeting`, produces
+the value of the `greeting` field. It is monadic so that it will only be
+executed when the field was requested.
+
+Each handler is a separate monadic action so we only perform the side effects
```
teh: Each handler -> Each field handler?

docs/source/tutorial/Introduction.lhs
```diff
@@ -12,63 +12,146 @@ module Introduction where
...
+### Running queries
+
+Defining a service isn't much point unless you can query. Here's how:
+
+```haskell
+queryHello :: IO Response
+queryHello = interpretAnonymousQuery @Hello hello "{ greeting(who: \"mort\") }"
+```
 
+The actual `Response` type is fairly big, so we're most likely to turn it into
```
teh: big -> verbose?

docs/source/tutorial/Introduction.lhs
```diff
@@ -77,107 +160,142 @@ type Calculator = Object "Calculator" '[]
...
 ```
 
-Note that, while `unionValue` looks a bit like `unsafeCoerce` by
-forcing one type to become another type, it's actually type-safe
-because we use a *type-index* to pick the correct type from the
-union. Using e.g. `unionValue @HelloWorld handler` will not compile
-because `HelloWorld` is not in the union.
-
-## Running a query
+And let's define a very simple top-level object that uses `UserOrCalcualtor`:
```
teh: UserOrCalculator (spelling)

docs/source/tutorial/Introduction.lhs
```diff
@@ -77,107 +160,142 @@ type Calculator = Object "Calculator" '[]
...
 
 ```haskell
-hello :: IO Response
-hello = interpretAnonymousQuery @HelloWorld handler "{ me(greeting: \"hey ho\") }"
+type UnionQuery = Object "UnionQuery" '[] '[Field "union" UserOrCalcualtor]
```
teh: same, fix spelling everywhere

PR #91, opened 2017-01-27
Wai Application adpater.

teh:


graphql-wai/src/GraphQL/Wai.hs
```diff
@@ -0,0 +1,35 @@
...
+-- in time.
+--
+-- If you have a 'Cat' type and a corresponding 'catHandler' then you
+-- can use "toApplication @Cat catHandler".
+toApplication :: forall r. (HasResolver IO r) => Handler IO r -> Application
+toApplication handler = app
+  where
+    app req respond =
+      case queryString req of
+        [("query", Just query)] -> do
```
jml: Maybe also handle POST requests, as per http://graphql.org/learn/serving-over-http/#post-request?

teh: Hm - IMO a server should be either GET or POST, not both, as that introduces more things that can wrong. What do you think about adding `toPostApplication` to handle POST, leaving the policy decision to the implementer?

jml: I guess it feels a bit off for me for our opinion to override the one recommended in the official GraphQL documentation, which is surely based on experience. 

However, no intrinsic objection to providing two separate things. Happier if they can be combined into one handler, but that's not necessary. As you've pointed out elsewhere, there's not much to this, let's get it merged and then iterate.

stack.yaml
```diff
@@ -3,3 +3,4 @@ resolver: nightly-2017-01-25
 packages:
   - "."
   - "./docs/source/tutorial"
+  - "./graphql-wai"
```
jml: Hmm.  I guess we should move to a mono-repo structure officially. Not in this PR.

graphql-wai/src/GraphQL/Wai.hs
teh: I don't think the docs say you should handle both, the way I read them is "if it's a GET, here's how you do it, if it's a POST, here's how you do it"

stack.yaml
teh: How would a mono-repo be different? Move top-level stuff to sub-dirs? OK with that but year, separate PR

jml: Yes, that.

teh:
Will merge. I want more stuff here but not yet. As discussed I think adding too much now, and deciding policies for our users at this stage is counterproductive.

PR #92, opened 2017-01-28
Merge selections in a set into ordered map of fields

### LINKED ISSUE #59: Enforce field uniqueness in selections

jml:
Fixes #59 

This makes the job of the resolver much simpler. It doesn't have to worry about inline fragments or whatever, it only deals with ordered maps of fields. 

Unfortunately, it also makes the job of the validator more complex. Recursing through the whole query to merge fields is a bit involved. Defining some generic operations on `OrderedMap` made this much easier to think about.

This PR also introduces the notion of types into the validator. We need to know the type of the object we're building a selection set for, and we need to know the definitions of the type conditions. This (to me) means building and providing a value that represents the whole schema, and making sure each resolver has it. This in turn means passing a new value, `allTypes` down through each resolver. I think this is actually a perfect use case for `ReaderT`, and would be happy to change it.

It also means changing the top-level methods to only accept GraphQL Objects. This is fine, because that's all their allowed to take: https://facebook.github.io/graphql/#sec-Initial-types

Something that naturally arises from this work is that the difference between leafs and non-leafs becomes really obvious. We talked in person about different approaches, and for this PR I settled on changing the type of resolver to take `Maybe (SelectionSet Value)`, where `Nothing` indicates a leaf query.

Testing is pretty shallow. I added one end-to-end test that shows off what this can do, but it's kind of limited. Some of the algorithmic & traversal code could probably do with property tests. Let me know which bits you trust least.

Some things on which I'd particularly like input:

- The name of `ExecutionField` (and names of new functions & types in general)
- Whether or not to use `ReaderT`
- Ways to reduce the implementation complexity of `groupByResponseKey`

### DIFF
--- a/src/GraphQL/Internal/Schema.hs
+++ b/src/GraphQL/Internal/Schema.hs
@@ -34,6 +34,10 @@ module GraphQL.Internal.Schema
   , NonNullType(..)
   , DefinesTypes(..)
   , doesFragmentTypeApply
+  -- * The schema
+  , Schema
+  , makeSchema
+  , lookupType
   ) where
 
 import Protolude hiding (Type)
@@ -41,6 +45,23 @@ import Protolude hiding (Type)
 import qualified Data.Map as Map
 import GraphQL.Value (Value)
 import GraphQL.Internal.Name (HasName(..), Name, unsafeMakeName)
+
+-- | An entire GraphQL schema.
+--
+-- This is very much a work in progress. Currently, the only thing we provide
+-- is a dictionary mapping type names to their definitions.
+newtype Schema = Schema (Map Name TypeDefinition) deriving (Eq, Show)
+
+-- | Create a schema from the root object.
+--
+-- This is technically an insufficient API, since not all types in a schema
+-- need to be reachable from a single root object. However, it's a start.
+makeSchema :: ObjectTypeDefinition -> Schema
+makeSchema = Schema . getDefinedTypes
+
+-- | Find the type with the given name in the schema.
+lookupType :: Schema -> Name -> Maybe TypeDefinition
+lookupType (Schema schema) name = Map.lookup name schema
 
 -- XXX: Use the built-in NonEmptyList in Haskell
 newtype NonEmptyList a = NonEmptyList [a] deriving (Eq, Show, Functor, Foldable)
--- a/src/GraphQL/Internal/Schema.hs
+++ b/src/GraphQL/Internal/Schema.hs
@@ -32,6 +32,8 @@ module GraphQL.Internal.Schema
   , AnnotatedType(..)
   , ListType(..)
   , NonNullType(..)
+  , DefinesTypes(..)
+  , doesFragmentTypeApply
   ) where
 
 import Protolude hiding (Type)
@@ -196,3 +198,28 @@ instance HasName InputTypeDefinition where
 -- 'InputObjectFieldDefinition') can have default values. These are allowed to
 -- be any kind of literal.
 type DefaultValue = Value
+
+
+-- | Does the given object type match the given type condition.
+--
+-- See <https://facebook.github.io/graphql/#sec-Field-Collection>
+--
+-- @
+-- DoesFragmentTypeApply(objectType, fragmentType)
+--   If fragmentType is an Object Type:
+--     if objectType and fragmentType are the same type, return true, otherwise return false.
+--   If fragmentType is an Interface Type:
+--     if objectType is an implementation of fragmentType, return true otherwise return false.
+--   If fragmentType is a Union:
+--     if objectType is a possible type of fragmentType, return true otherwise return false.
+-- @
+doesFragmentTypeApply :: ObjectTypeDefinition -> TypeDefinition -> Bool
+doesFragmentTypeApply objectType fragmentType =
+  case fragmentType of
+    TypeDefinitionObject obj -> obj == objectType
+    TypeDefinitionInterface interface -> objectType `implements` interface
+    TypeDefinitionUnion union -> objectType `branchOf` union
+    _ -> False
+  where
+    implements (ObjectTypeDefinition _ interfaces _) int = int `elem` interfaces
+    branchOf obj (UnionTypeDefinition _ (NonEmptyList branches)) = obj `elem` branches
--- a/src/GraphQL/Internal/Schema.hs
+++ b/src/GraphQL/Internal/Schema.hs
@@ -50,7 +50,7 @@ import GraphQL.Internal.Name (HasName(..), Name, unsafeMakeName)
 --
 -- This is very much a work in progress. Currently, the only thing we provide
 -- is a dictionary mapping type names to their definitions.
-newtype Schema = Schema (Map Name TypeDefinition) deriving (Eq, Show)
+newtype Schema = Schema (Map Name TypeDefinition) deriving (Eq, Ord, Show)
 
 -- | Create a schema from the root object.
 --
@@ -64,7 +64,7 @@ lookupType :: Schema -> Name -> Maybe TypeDefinition
 lookupType (Schema schema) name = Map.lookup name schema
 
 -- XXX: Use the built-in NonEmptyList in Haskell
-newtype NonEmptyList a = NonEmptyList [a] deriving (Eq, Show, Functor, Foldable)
+newtype NonEmptyList a = NonEmptyList [a] deriving (Eq, Ord, Show, Functor, Foldable)
 
 -- | A thing that defines types. Excludes definitions of input types.
 class DefinesTypes t where
@@ -82,7 +82,7 @@ class DefinesTypes t where
 data AnnotatedType t = TypeNamed t
                      | TypeList (ListType t)
                      | TypeNonNull (NonNullType t)
-                     deriving (Eq,Show)
+                     deriving (Eq, Ord, Show)
 
 -- | Get the type that is being annotated.
 getAnnotatedType :: AnnotatedType t -> t
@@ -94,13 +94,13 @@ getAnnotatedType (TypeNonNull (NonNullTypeList (ListType t))) = getAnnotatedType
 instance HasName t => HasName (AnnotatedType t) where
   getName = getName . getAnnotatedType
 
-newtype ListType t = ListType (AnnotatedType t) deriving (Eq, Show)
+newtype ListType t = ListType (AnnotatedType t) deriving (Eq, Ord, Show)
 
 data NonNullType t = NonNullTypeNamed t
                    | NonNullTypeList  (ListType t)
-                   deriving (Eq,Show)
+                   deriving (Eq, Ord, Show)
 
-data Type = DefinedType TypeDefinition | BuiltinType Builtin deriving (Eq, Show)
+data Type = DefinedType TypeDefinition | BuiltinType Builtin deriving (Eq, Ord, Show)
 
 instance DefinesTypes Type where
   getDefinedTypes (BuiltinType _) = mempty
@@ -117,7 +117,7 @@ data TypeDefinition = TypeDefinitionObject        ObjectTypeDefinition
                     | TypeDefinitionEnum          EnumTypeDefinition
                     | TypeDefinitionInputObject   InputObjectTypeDefinition
                     | TypeDefinitionTypeExtension TypeExtensionDefinition
-                      deriving (Eq, Show)
+                      deriving (Eq, Ord, Show)
 
 instance HasName TypeDefinition where
   getName (TypeDefinitionObject x) = getName x
@@ -141,7 +141,7 @@ instance DefinesTypes TypeDefinition where
         panic "TODO: we should remove the 'extend' behaviour entirely"
 
 data ObjectTypeDefinition = ObjectTypeDefinition Name Interfaces (NonEmptyList FieldDefinition)
-                            deriving (Eq, Show)
+                            deriving (Eq, Ord, Show)
 
 instance HasName ObjectTypeDefinition where
   getName (ObjectTypeDefinition name _ _) = name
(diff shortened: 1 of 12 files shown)


src/GraphQL/Resolver.hs
```diff
@@ -134,17 +133,24 @@ data Result a = Result [ResolverError] a deriving (Show, Functor, Eq)
 aggregateResults :: [Result Value] -> Result Value
 aggregateResults r = toValue <$> sequenceA r
 
+throwE :: Applicative f => ResolverError -> f (Result Value)
```
teh: is overloading the well-known `throwE` name a good idea?

src/GraphQL/Resolver.hs
```diff
@@ -134,17 +133,24 @@ data Result a = Result [ResolverError] a deriving (Show, Functor, Eq)
...
+throwE err = pure (Result [err] GValue.ValueNull)
+
 instance Applicative Result where
   pure v = Result [] v
   (Result e1 f) <*> (Result e2 x) = Result (e1 <> e2) (f x)
 
 ok :: Value -> Result Value
 ok = pure
 
+type AllTypes = Map Name TypeDefinition
```
teh: Can you add a comment about `AllTypes` intent? Not super sure from the code.

src/GraphQL/Resolver.hs
```diff
@@ -379,30 +382,33 @@ instance forall m a b dispatchType.
...
-    let (Result errs obj)  = GValue.makeObject . catMaybes <$> sequenceA r
-    case obj of
-      Nothing -> pure (Result [DuplicateFields r] GValue.ValueNull)
-      Just object -> pure (Result errs (GValue.ValueObject object))
+  resolve _ _ Nothing = throwE MissingSelectionSet
+  resolve allTypes mHandler (Just selectionSet) =
+    case getSelectionSet of
+      Left err -> throwE err
+      Right ss -> do
+        handler <- mHandler
```
teh: Can you re-add the a comment along the lines of "First we run the handler function itself." ? Phrase how you want it but I find communicating that intent here quite important.

src/GraphQL/Resolver.hs
```diff
@@ -134,17 +133,24 @@ data Result a = Result [ResolverError] a deriving (Show, Functor, Eq)
...
 
+type AllTypes = Map Name TypeDefinition
 
 class HasResolver m a where
   type Handler m a
-  resolve :: Handler m a -> SelectionSet Value -> m (Result Value)
+  -- XXX: Might be nicer to make AllTypes a ReaderT -- easier to ignore when
+  -- not wanted, and wouldn't have required me to change so many call sites.
+  -- OTOH, with all our type weirdness, I got bored trying to figure it out.
+  resolve :: AllTypes -> Handler m a -> Maybe (SelectionSetByType Value) -> m (Result Value)
```
teh: I don't have an answer here but adding a new argument to `HasResolver` is not good from an API POV IMO. This now means that our users have to carry around this extra argument which is an implementation detail every time they implement `HasResolver`.

src/GraphQL/Internal/Schema.hs
```diff
@@ -32,21 +32,47 @@ module GraphQL.Internal.Schema
...
+import qualified Data.Map as Map
 import GraphQL.Value (Value)
 import GraphQL.Internal.Name (HasName(..), Name, unsafeMakeName)
 
 -- XXX: Use the built-in NonEmptyList in Haskell
-newtype NonEmptyList a = NonEmptyList [a] deriving (Eq, Show)
+newtype NonEmptyList a = NonEmptyList [a] deriving (Eq, Show, Functor, Foldable)
+
+-- | A thing that defines types. Excludes definitions of input types.
+class DefinesTypes t where
```
teh: No `HasX` naming schema?

src/GraphQL/Resolver.hs
```diff
@@ -379,30 +382,33 @@ instance forall m a b dispatchType.
...
+      Right ss -> do
+        handler <- mHandler
+        r <- traverse (runFields @m @(RunFieldsType m fields) allTypes handler) ss
+        let (Result errs obj)  = GValue.objectFromOrderedMap . OrderedMap.catMaybes <$> sequenceA r
+        pure (Result errs (GValue.ValueObject obj))
 
+    where
+      getSelectionSet = do
+        defn <- first SchemaError $ API.getDefinition @(API.Object typeName interfaces fields)
+        (SelectionSet ss') <- first ValidationError $ getSelectionSetForType (flip Map.lookup allTypes) defn selectionSet
```
teh: Can you add a comment why we need the lookup in `allTypes`?

src/GraphQL/Resolver.hs
jml: We already do this in Validation.hs. Happy to rename, but totally bored of rewriting the same `Result` expression a dozen times over.

src/GraphQL/Resolver.hs
jml: Something like:

```haskell
-- | Dictionary of all types in the schema. A map of type names to type definitions.
```

?

src/GraphQL/Resolver.hs
jml: I'd much rather put in a comment saying why we run the handler first. Why do we run the handler first?

src/GraphQL/Resolver.hs
jml: Some reflections. These are mostly unrelated to each other, just a bunch of points loosely on topic.

* What's the `resolve` API for from a users PoV? I genuinely don't know. Knowing that seems the _sine qua non_ in a discussion about usability.
* This is essentially a callback API. It's pretty standard for callbacks to provide arguments that are unused in many cases. We already (before this PR) have many instances that don't use arguments.
* Where callbacks _don't_ provide arguments that are unused, it's because there are two different callback points. This would equate to two different type classes, or two different methods on this one.
* `AllTypes` is something of a stand-in for a value that represents the entire global schema, not just the element that's being resolved. This makes it something more than just an implementation detail.
* If a user _did_ want to write a resolver that had access to the entire schema (as I did), what should they do?
* `ReaderT` is a pretty good compromise here, IMO, since it can be safely ignored by those who don't need it. It also makes it harder to get it wrong by passing a different value into a child resolver (which we alas have to do)
* With all the type magic going on, do you really think an extra argument is the critical usability point?
* If we intend people to provide implementations of `HasResolver`, we should document the class and its methods, and say what the contract of the implementations is.

src/GraphQL/Internal/Validation.hs
```diff
@@ -208,18 +192,165 @@ validateOperation (Mutation vars directives selectionSet) = do
...
+
+newtype SelectionSetByType value
+  = SelectionSetByType (OrderedMap ResponseKey (OrderedMap (Set TypeCondition) (ExecutionField value)))
+  deriving (Eq, Ord, Show, Functor, Foldable, Traversable)
+
+-- | A 'ResponseKey' is the key under which a field appears in a response. If
+-- there's an alias, it's the alias, if not, it's the field name.
+type ResponseKey = Name
+
+-- | A field ready to be executed.
```
jml: Sure. You'd rather "resolved"?

src/GraphQL/Internal/Validation.hs
```diff
@@ -208,18 +192,165 @@ validateOperation (Mutation vars directives selectionSet) = do
...
+-- XXX: Reviewer, the 'Field' name has been freed up. Should I rename this to 'Field'.
```
jml: OK, will do.

src/GraphQL/Internal/Validation.hs
```diff
@@ -208,18 +192,165 @@ validateOperation (Mutation vars directives selectionSet) = do
...
+                          , arguments = arguments field1
+                          , subSelectionSet = Nothing
+                          }
+    (Just ss1, Just ss2) -> do
+      mergedSet <- mergeSelectionSets ss1 ss2
+      pure ExecutionField { name = name field1
+                          , arguments = arguments field1
+                          , subSelectionSet = Just mergedSet
+                          }
+    _ -> throwE (IncompatibleFields (name field1))
```
jml: Nope, it's what happens when someone tries:

```graphql
{
  dog {
    barkVolume
  }
  dog
}
```

src/GraphQL/Internal/Schema.hs
jml: I actually thought our naming schema was "predicate phrase", but sure. `HasTypeDefinitions`?

src/GraphQL/Resolver.hs
jml: Thinking about it, a different solution would be to change `Validation` to require a type dictionary (and later, a validated schema), and then substitute the type conditions for the actual types.

src/GraphQL/Resolver.hs
jml: Something like...

```haskell
-- Fields of a selection set may be behind "type conditions", due to inline fragments 
-- or the use of fragment spreads. These type conditions are represented in the 
-- schema by the name of a type (e.g. "Dog"). To determine which type conditions 
-- (and thus which fields) are relevant for this selection set, we need to look up the 
-- actual types they refer to, as interfaces (say) match objects differently than 
-- unions.
-- 
-- See <https://facebook.github.io/graphql/#sec-Field-Collection> for more details.
```

Can you please provide similar levels of commentary on literally everything in Resolver.hs?

jml:
> This changes so much that I have trouble finding my way in. Continuing but some preliminary questions.

Thanks for making the effort. I realise it's pretty big, but (aside from the OrderedMap changes) I couldn't break it up into smaller pieces without either having chunks of unused code, or making changes that would seem to be without any justification.

src/GraphQL/Internal/Schema.hs
teh: ah right, not sure.  We have a mix, can't really place it anywhere so probably fine

```
$ gg -h "^class " | sort | uniq
class BuildFieldResolver m fieldResolverType where
class Defaultable a where
class FromValue a where
class GenericAnnotatedInputType (f :: Type -> Type) where
class GenericEnumValues (f :: Type -> Type) where
class GenericFromValue (f :: Type -> Type) where
class GenericInputObjectFieldDefinitions (f :: Type -> Type) where
class GraphQLEnum a where
class GraphQLError e where
class HasAnnotatedInputType a where
class HasAnnotatedType a where
class HasArgumentDefinition a where
class HasFieldDefinition a where
class HasFieldDefinitions a where
class HasInterfaceDefinition a where
class HasInterfaceDefinitions a where
class HasName a where
class HasObjectDefinition a where
class HasResolver m a where
class RunFields m a where
class RunUnion m union objects where
class ToValue a where
class UnionTypeObjectTypeDefinitionList a where
```

src/GraphQL/Internal/Validation.hs
teh: Ah I see, how annoying! I wonder what a query language looks like that doesn't allow this .. probably doesn't have fragments!

src/GraphQL/Internal/Validation.hs
jml: Just came across https://facebook.github.io/graphql/#sec-Executing-Fields

src/GraphQL/Resolver.hs
teh: All interesting points!

The API (as envisioned by my so far) would be (in the simple case) that I have a new type, e.g.

```haskell
data BusinessLogicThing = MoneyYes { amount :: Int } deriving Generic
```

and then 

```haskell
instance HasResolver BusinessLogicThing where
   resolve handler _ = map businessLogicThingToValue handler
```

The more complex (and probably more interesting) case is that users implement new combinators, e.g. `ReadonlyObject` or `DynamicSchemaObject` (the latter could e.g. figure out what fields it supports on the fly). 

I think at this stage I agree with `AllTypes` being a stand-in for the full schema, and think we should keep it.

I'm also capitulating in terms of design, I really need some larger-project feedback now before making any other decisions.

Summary: :shipit:

src/GraphQL/Resolver.hs
teh: Ah right. The idea is that we want to run some monadic action that then determines what goes into the fields. E.g. we could be opening a database connection, or delegating to some other service.

src/GraphQL/Resolver.hs
teh: E.g. one thing that I imagine could be be quite useful is to delegate to a different graphql endpoint in the system, like `DelegationObject`, i.e. we would not know in advance what to do, and we'd need to re-bundle the query and ship it to the other place (maybe giving it a new name along the way).

src/GraphQL/Internal/Validation.hs
jml: Well, it *doesn't* allow it, that's why we're catching it in validation. We just can't reasonably test the pairings without knowing the types, and we're (contra spec) not doing type-level validation atm. In the next major release (if we get there), I expect that the introspection work and pre-query type-checking will go hand-in-hand, much as the resolver & validation work did in this release.

src/GraphQL/Resolver.hs
jml: Cool. I'll make `AllTypes` a `newtype` with only a lookup operation so as to make this a bit more obvious.

src/GraphQL/Internal/Schema.hs
jml: `BuildFieldResolver` and `UnionTypeObjectTypeDefinitionList` really stand out in that list.

src/GraphQL/Resolver.hs
teh: ok, makes sense

src/GraphQL/Resolver.hs
jml: Sorry, am still struggling to figure out why this needs to be written down. It feels like writing "do `a -> m b` so we can pass `b` to the next thing".

Wrote this:

```haskell
-- Run the handler so the field resolvers have access to the object.
-- This (and other places, including field resolvers) is where user
-- code can do things like look up something in a database.
```

PR #117, opened 2017-03-20
I116

teh:


docs/requirements.txt
```diff
@@ -1,25 +0,0 @@
-alabaster==0.7.7
-argh==0.26.1
```
jml: Did you mean to remove this file as part of this change?

teh: Yea, as discussed in #114 this was an attempt at making syntax highlighting work, and it didn't work, so removing again.

Issue #119, opened 2017-04-11
Question: mapping over a Handler?


lexi-lambda:
*This is a reproduction of an open [Stack Overflow question](http://stackoverflow.com/q/43333486/465378), which I think might get more visibility here.*

I have created a simple API for testing purposes:

```haskell
type Query = Object "Query" '[]
  '[ Argument "id" Text :> Field "test" (Maybe Foo) ]

type Foo = Object "Foo" '[]
  '[ Field "name" Text ]
```

I also have a Haskell type that represents the `Foo` resource, as well as a function for retrieving it from the database:

```haskell
data ServerFoo = ServerFoo
  { name :: Text
  } deriving (Eq, Show)

lookupFoo :: Text -> IO (Maybe ServerFoo)
```

Now, I want to implement a `Handler` for `Query`. To start, I implemented a `Handler` for `Foo`:

```haskell
viewFoo :: ServerFoo -> Handler IO Foo
viewFoo ServerFoo { name } = pure $ pure name
```

Then I went to implement the root handler, but I realized I am not sure how to use my `viewFoo` function when I end up with with a `Maybe ServerFoo`. I tried this:

```haskell
handler :: Handler IO Query
handler = pure $ \fooId -> do
  foo <- lookupFoo fooId
  sequence $ fmap viewFoo foo
```

However, this does not work. It produces the following type error:

    • Couldn't match type ‘IO Text’
                     with ‘Object "Foo" '[] '[Field "name" Text]’
      Expected type: ServerFoo
                     -> IO (Object "Foo" '[] '[Field "name" Text])
        Actual type: ServerFoo -> Handler IO Foo
    • In the first argument of ‘fmap’, namely ‘viewFoo’
      In the second argument of ‘($)’, namely ‘fmap viewFoo foo’
      In a stmt of a 'do' block: (sequence $ fmap viewFoo foo)

This seems a bit odd to me, given that the examples I’ve read would seem to imply that `Handler`s are designed to be compositional. Indeed, they appear to be, since making a few tweaks to remove the `Maybe` makes this typecheck:

```haskell
type Query = Object "Query" '[]
  '[ Argument "id" Text :> Field "test" Foo ]

handler :: Handler IO Query
handler = pure $ \fooId -> do
  Just foo <- lookupFoo fooId
  viewFoo foo
```

Obviously, however, this involves a partial pattern match, so it’s not an okay solution.

[...]

PR #120, opened 2017-04-18
Correctly propagate Maybe handlers. Closes #102 and #119.

### LINKED ISSUE #102: Nullable handler doesn't support objects
Probably.

Code is:

```haskell
instance forall m hg. (HasResolver m hg, Functor m, ToValue (Maybe hg)) => HasResolver m (Maybe hg) where
  type Handler m (Maybe hg) = m (Maybe hg)
  resolve handler _ =  map (ok . toValue) handler
```

Note that it discards the selection set. That can't be good.

teh:
I'm not happy with this solution because of the extra monad level
introduced.

    pure (Just (pure value))

is not a good way to return a Maybe. Not sure how to fix better ATM.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,10 +3,19 @@ graphql-api changelog
 =====================
 
 $next (yyyy-mm-dd)
-==================
+=================
+
+
+v0.2.0 (2017-10-12)
+===================
 
 * Make ``Name`` an overloaded string that panics if an invalid name is
   provided.
+* Correctly descend into the type parameter of a ``Maybe``. See https://github.com/jml/graphql-api/issues/119.
+  This is a backwards-incompatible change.
+
+  A common update would be having to ``fmap pure callback`` instead of just ``callback``
+  for ``Maybe`` handlers.
 
 
 v0.1.0 (2017-01-30)
@@ -21,5 +30,3 @@ v0.1.0 (2017-01-29)
 ===================
 
 Initial release, support basic handling of GraphQL queries.
-
-
--- a/tests/ResolverTests.hs
+++ b/tests/ResolverTests.hs
@@ -37,6 +37,33 @@ tHandler :: Handler TMonad T
 tHandler =
   pure $ (pure 10) :<> (\tArg -> pure tArg) :<> (pure . (*2))
 
+
+-- https://github.com/jml/graphql-api/issues/119
+-- Maybe X didn't descend into its argument. Now it does.
+type Query = Object "Query" '[]
+  '[ Argument "id" Text :> Field "test" (Maybe Foo) ]
+
+type Foo = Object "Foo" '[]
+  '[ Field "name" Text ]
+
+data ServerFoo = ServerFoo
+  { name :: Text
+  } deriving (Eq, Show)
+
+lookupFoo :: Text -> IO (Maybe ServerFoo)
+lookupFoo _ = pure $ Just (ServerFoo "Mort")
+
+viewFoo :: ServerFoo -> Handler IO Foo
+viewFoo ServerFoo { name=name } = pure $ pure $ name
+
+handler :: Handler IO Query
+handler = pure $ \fooId -> do
+  foo <- lookupFoo fooId
+  -- note that fmap maps over the Maybe, so we still need
+  -- have to wrap the result in a pure.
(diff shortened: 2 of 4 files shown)


jml:
I don't like the extra monad level either. I'm not sure how to fix and I don't think I fully understand what's going on. OK to hold off a little to give us both time to think of something better?

teh:
The extra level happens because every "type translation" step in form of an `instance HasResolver` introduces a new monad level ATM.
```
  type Handler m (Maybe hg) = m (Maybe (Handler m hg))
                                        ^^^^^ this part expands to e.g. `m Text`
```

We can have arbitrary type computations, so we could e.g. add another type function to `HasResolver`:

```haskell
class HasResolver (m :: Type -> Type) (a :: Type) where
  type Handler m a
  type NonMonadicHandler a
  type instance NonMonadicHandler a = a
  resolve :: Handler m a -> Maybe (SelectionSetByType Value) -> m (Result Value)
```

and then use that to avoid introducing the extra `m`. This makes it slightly harder to reason about types but practically it might be more intuitive.

teh:
Ah, no, we still need a way of patching through `m`, and also `NonMonadicHandler` doesn't work when we expect `Handler` as input to `resolve`. I'll ponder.

teh:
UX-wise it'd be nicest if we could return maybes directly, e.g. `pure (Just "hello")` but then we can't use `Handler` type expansion because that introduces the extra `m` (and it's needed in the case that we return an object, not just a single `Text` field).

So just to keep composition I don't think we'll get around `Just (pure "hello")`, no matter how ugly.

A 2nd option is trying to peel off the outer `m`, so instead of `pure (Just (pure "hello"))` we'd have just `Just (pure "hello")`, but then we need to figure out a way of running the handler in `m`, i.e. what is `resolve` in the following?:

```haskell
instance forall m hg. (HasResolver m hg, Monad m) => HasResolver m (Maybe hg) where
  type Handler m (Maybe hg) = Maybe (Handler m hg)
  resolve handler selectionSet = ???
```

Slight tangent: We discussed a while back having a `PlainField` type that doesn't run in `m`, maybe that'd be a way of cushioning the API UX issues.

CHANGELOG.rst
```diff
@@ -3,10 +3,16 @@ graphql-api changelog
...
-==================
+=================
+
+
+v0.2.0 (2017-10-12)
+===================
 
 * Make ``Name`` an overloaded string that panics if an invalid name is
   provided.
+* Correctly descent into the type parameter of a `Maybe`. See https://github.com/jml/graphql-api/issues/119.
```
jml: `descend`

CHANGELOG.rst
```diff
@@ -3,10 +3,16 @@ graphql-api changelog
...
+  This is a backwards-incompatible change.
```
jml: Can you give an example of what would need to be updated?

tests/ResolverTests.hs
```diff
@@ -37,6 +37,33 @@ tHandler :: Handler TMonad T
 tHandler =
   pure $ (pure 10) :<> (\tArg -> pure tArg) :<> (pure . (*2))
 
+
+-- https://github.com/jml/graphql-api/issues/119
+-- Maybe X didn't descent into its argument. Now it does.
```
jml: `descend`

tests/ResolverTests.hs
```diff
@@ -37,6 +37,33 @@ tHandler :: Handler TMonad T
...
+
+viewFoo :: ServerFoo -> Handler IO Foo
+viewFoo ServerFoo { name=name } = pure $ pure $ name
+
+handler :: Handler IO Query
+handler = pure $ \fooId -> do
+  foo <- lookupFoo fooId
+  -- note that fmap maps over the Maybe, so we still need
+  -- have to wrap the result in a pure.
+  sequence $ fmap (pure . viewFoo) foo
```
jml: A little bit weird that this isn't executed. I guess we're relying on the compiler?

tests/ResolverTests.hs
jml: Still needs fixing.

teh: done

tests/ResolverTests.hs
teh: fixed

PR #123, opened 2017-09-21
Update to protolude 0.2

brynedwards:
Hi,
this should update graphql-api to compile with protolude 0.2 on both GHC 8.0.2 and 8.2.1.


jml:
Thanks!

PR #128, opened 2017-11-19
Create a Request object and an interpretRequest path

dmaze:
There is a mostly-standard [JSON request format](http://graphql.org/learn/serving-over-http/#post-request) that's pretty widely support across GraphQL implementations, even if not part of the spec _per se_.  This PR adds a `Request` type that mirrors that format, and a top-level `interpretRequest` entry point to run it.

The real goal of this is to allow a Servant path like

```haskell
type GraphQLAPI = "graphql"
                  :> ReqBody '[JSON] GraphQL.Request
                  :> Post '[JSON] GraphQL.Response

graphQLServer :: Server GraphQLAPI
graphQLServer = liftIO . runRequest
```

The one big caveat in this PR is that you can't 100% deserialize `Value` objects without knowing their type context.  This assumes that JSON strings are always GraphQL strings (they could also be enums or schema-specific scalar types) and that JSON numbers are always GraphQL floats (they could also be integers and that's probably the common case).

### DIFF
--- a/src/GraphQL/Internal/Syntax/AST.hs
+++ b/src/GraphQL/Internal/Syntax/AST.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -104,6 +105,12 @@ instance IsString Name where
 instance Aeson.ToJSON Name where
   toJSON = Aeson.toJSON . unName
 
+instance Aeson.FromJSON Name where
+  parseJSON = Aeson.withText "Name" $ \v ->
+    case makeName v of
+      Left _ -> mempty
+      Right name -> return name
+
 instance Arbitrary Name where
   arbitrary = do
     initial <- elements alpha
@@ -153,7 +160,11 @@ getNodeName (Node name _ _ _) = name
 data VariableDefinition = VariableDefinition Variable Type (Maybe DefaultValue)
                           deriving (Eq,Show)
 
-newtype Variable = Variable Name deriving (Eq, Ord, Show)
+newtype Variable = Variable Name deriving (Eq, Ord, Show, Aeson.FromJSON,
+                                           Aeson.ToJSON)
+
+instance Aeson.FromJSONKey Variable
+instance Aeson.ToJSONKey Variable
 
 instance Arbitrary Variable where
   arbitrary = Variable <$> arbitrary
--- a/src/GraphQL/Value.hs
+++ b/src/GraphQL/Value.hs
@@ -51,8 +51,11 @@ module GraphQL.Value
 import Protolude
 
 import qualified Data.Aeson as Aeson
-import Data.Aeson (ToJSON(..), (.=), pairs)
+import Data.Aeson (FromJSON(..), ToJSON(..), (.=), pairs)
+import qualified Data.HashMap.Lazy as HashMap
 import qualified Data.Map as Map
+import Data.Scientific (toRealFloat)
+import qualified Data.Vector as Vector
 import Test.QuickCheck (Arbitrary(..), Gen, oneof, listOf, sized)
 
 import GraphQL.Internal.Arbitrary (arbitraryText)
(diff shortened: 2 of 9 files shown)


src/GraphQL/Internal/Syntax/AST.hs
```diff
@@ -104,6 +105,12 @@ instance IsString Name where
 instance Aeson.ToJSON Name where
   toJSON = Aeson.toJSON . unName
 
+instance Aeson.FromJSON Name where
+  parseJSON = Aeson.withText "Name" $ \v ->
+    case makeName v of
+      Left _ -> mempty
```
teh: Is `mempty` the right thing here? Should we maybe `fail` instead? I'm not even sure why `mempty` is a valid value for `Name` - maybe @jml remembers :)

src/GraphQL/Value.hs
```diff
@@ -160,6 +173,13 @@ data ConstScalar
   | ConstNull
   deriving (Eq, Ord, Show)
 
+instance FromJSON ConstScalar where
+  parseJSON (Aeson.String x) = parseJSON (Aeson.String x) >>= return . ConstString
+  parseJSON (Aeson.Number x) = return $ ConstFloat $ toRealFloat x
+  parseJSON (Aeson.Bool x) = return $ ConstBoolean x
+  parseJSON Aeson.Null = return ConstNull
+  parseJSON _ = mempty
```
teh: What's the `mempty` for `ConstScalar`?

src/GraphQL/Internal/Syntax/AST.hs
dmaze: In both this and the other case, you're [running inside the Aeson `Parser` monad](https://hackage.haskell.org/package/aeson-1.2.3.0/docs/Data-Aeson.html#t:FromJSON).  Protolude is a little more aggressive about moving `fail` out of the base `Monad` class, so it looks like I need to `import Control.Monad.Fail` and then it works.

jml: @teh This is a wart in Aeson IMO. Because `mempty` wasn't prefixed with `return` (or `pure`), it's `mempty :: Aeson.Parser Name`, not `mempty :: Name` (which wouldn't compile).

If you look at the source for `Parser a`, it has [an implementation of `Monoid`](http://hackage.haskell.org/package/aeson-1.2.3.0/docs/src/Data-Aeson-Types-Internal.html#line-323) such that `mempty = fail "mempty"`. I think this kind of sucks.

`empty` (i.e. failed `Alternative`, not monoid identity), would be much more correct.

But `fail` is best, because it's informative.

src/GraphQL/Value.hs
jml: It's not `mempty` for `ConstScalar`, it's `mempty` for `Parser a`.

jml:
Hi @dmaze,

Thanks for this PR—it seems like a good idea. I'd like to have a look over it before merging to make sure I understand it.

Out of curiosity, can you share a little about what you're using graphql-api and servant for?

Thanks,
jml

dmaze:
My day job is using GraphQL pretty successfully, and in particular, the React+Relay+Flow JavaScript stack has been a pretty comfortable way to write front-end applications that don't need much state beyond what they can get from the server.  For my personal tasks I prefer to use languages where, say, it's _possible_ for my editor to notice simple typos.

The specific thing I'm actually playing with is an application that reads the MBTA's (Boston public transit agency) [real-time data feeds](https://realtime.mbta.com/), caches them in a Persistent (SQLite) database, then re-serves the cached data to a front-end that can display the actual travel times between a pair of stops on a day-to-day basis (while the T _claims_ the Red Line has a 92% on-time rate, it feels to my like at least 25% of my commute-time trips have significant delays).  If the back-end exposed a GraphQL interface, then the front-end would be very familiar space to me...which brought me here.

src/GraphQL/Value.hs
```diff
@@ -302,6 +325,16 @@ objectFromList xs = Object' <$> OrderedMap.orderedMap xs
 unionObjects :: [Object' scalar] -> Maybe (Object' scalar)
 unionObjects objects = Object' <$> OrderedMap.unions [obj | Object' obj <- objects]
 
+instance FromJSON scalar => FromJSON (Object' scalar) where
+  parseJSON = Aeson.withObject "Object" $ \v -> do
+    -- Order of keys is lost before we get here
```
jml: Can you remind me why this is relevant?

src/GraphQL/Value.hs
```diff
@@ -151,6 +161,11 @@ toObject _ = empty
 -- * Scalars
 
 -- | A non-variable value which contains no other values.
+--
+-- Note that the 'FromJSON' instance always decodes JSON strings to
+-- GraphQL strings (never enums) and JSON numbers to GraphQL floats
+-- (never ints); doing a better job of resolving this requires query
+-- context.
```
jml: I'm a little uncomfortable about this. The "Zen of Python" advises "in the face of ambiguity, refuse the temptation to guess", and I think that's good advice. 

Some options I could think of:
- _don't_ have a `FromJSON` instance for this, and instead have a series of functions that take the query (or whatever the minimal necessary context is) and a `Value` to be parsed, so that we can unambiguously parse these things
- create a new type (types?) that unites the query context with these values and write `FromJSON` instances for these new types
- create a new `AmbiguousConstScalar` type. Then, have a function that transforms `AmbiguousConstScalar -> ConstScalar` by also taking a query context. This type would need fewer branches than `ConstScalar`.
- make `ConstScalar` a phantom type, where the phantom parameter is whether it's ambiguous or not. Then, have a function that transforms `ConstScalar Ambiguous -> ConstScalar Unambiguous` by also taking a query context

PR #134, opened 2017-12-16
Make Node instance of HasName. Implement getName.

sunwukonga:
Following todo formerly in src/GraphQL/Internal/Syntax/AST.hs:
  TODO: Just make Node implement HasName.
Declared Node as instance of HasName and wrote implementation of
getname for it.

Because of a cyclic dependency between Name and AST, moved the Name
specific code from GraphQL.Internal.Syntax.AST module into the
GraphQL.Internal.Name module.

Updated imports and exposures in the AST and Name modules described
above.

Simple import and qualified name changes to:
  GraphQL/Internal/Syntax/Encoder
  GraphQL/Internal/Syntax/Parser

### DIFF
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -50,7 +50,7 @@ unsafeMakeName name =
 -- | Create a 'Name'.
 --
 -- Names must match the regex @[_A-Za-z][_0-9A-Za-z]*@. If the given text does
--- not match, return Nothing.
+-- not match, return NameError.
 --
 -- >>> makeName "foo"
 -- Right (Name {unName = "foo"})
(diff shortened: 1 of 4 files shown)


src/GraphQL/Internal/Name.hs
```diff
@@ -17,13 +18,57 @@ module GraphQL.Internal.Name
...
+unsafeMakeName :: HasCallStack => Text -> Name
+unsafeMakeName name =
+  case makeName name of
+    Left e -> panic (show e)
+    Right n -> n
+
+-- | Create a 'Name'.
+--
+-- Names must match the regex @[_A-Za-z][_0-9A-Za-z]*@. If the given text does
+-- not match, return Nothing.
```
jml: I realise this is just moved from another file, but could you please change it to "return a NameError" rather than "return Nothing"?

sunwukonga: No problem :)

Issue #135, opened 2017-12-18
[Question] Can schema match the root nodes {"query" : "...", "operationName": ... ?


sunwukonga:
Before I go off and write Aeson FromJSON instances for the root nodes of the GraphQL specification (for POSTs with content type "application/json"), is there already a way to wrangle it out of the Object data type when writing the schema?

I want to follow the format given at [GraphQL Post Request](http://graphql.org/learn/serving-over-http/#post-request).

That is:

    {
      "query": "...",
      "operationName": "...",
      "variables": { "myVariable": "someValue", ... }
    }

where either `operationName`, namely, `query` or `mutation`, are nested under `"query":`.

Just making sure before I do it.

teh:
@sunwukonga https://github.com/jml/graphql-api/pull/128/files might be interesting. We haven't quite figured out yet what to do about the `FromJSON` for `ConstScalar` instance. As implemented in that PR it has some limitations, and because of that I'm a bit nervous about introducing a `FromJSON` instance that'll tie us down in what we can decode in the future.

I.e. I'd prefer a solution that's `interpretJSONRequest` and does the variable decoding based on the schema but I don't have a fully formed picture of what that would look like yet.

Does that make sense?

sunwukonga:
The schema _may_ not be necessary, so long as I have all of the possibilities straight in my head. The only variable types we can get into the Query or Mutation Node are:
```
String        : A UTF‐8 character sequence.
ID            : a string, I presume the same as above, but not explicitly stated.
Enum          : a string, as above. (Often all-caps, but not sure this is 100% consistent).

Int           : A signed 32‐bit integer.
Float         : A signed double-precision floating-point value.

Boolean       : true or false.
Input Type    : Nested {}, making it easy to distinguish.
```

I considered `Union`s, but they don't go in any argument, which seems to be a condition for variable use. The problem: without context, there is ambiguity between the first three, and separately, between `Int` and `Float`. 

However, with the `QueryDocument`, (particularly the selected `operationName` Node) we can distinguish instances of String and ID easily. Filtering down, all JSON strings that remain must be `Enum`s. NB: no attempt is made to validate at this stage, i.e. search through the `SchemaDocument` for `EnumTypeDefinition` `Name`'s that match our `Enum` input.

By the same mechanism, `Int` and `Float` can be easily distinguished from each other. And even `scalar Date` if it's implemented.

[...]

Issue #132, opened 2017-12-11
Example with variables


eugenk:
The [tests](https://github.com/jml/graphql-api/blob/2b64b65175a315cc8a5247aa519793f40d8b6058/tests/EndToEndTests.hs#L240-L290) use a variable map, but the variable is not passed as a parameter to a handler. Instead, [the handler already has the variable hard coded](https://github.com/jml/graphql-api/blob/2b64b65175a315cc8a5247aa519793f40d8b6058/tests/EndToEndTests.hs#L241). How can a variable [from the variable map](https://github.com/jml/graphql-api/blob/2b64b65175a315cc8a5247aa519793f40d8b6058/tests/EndToEndTests.hs#L278-L279) be passed as a value to the handler?

PR #137, opened 2017-12-20
Fix error on anonymous query

sunwukonga:
#### Incoming query: ####
```
"{\"query\":\"query {\\n  greeting(who: \\\"Tim\\\")\\n}\"}"
```
Notice that the query (immediately after the start of the JSON field `query:`) has no operationName, i.e. it's anonymous.
#### This gets decoded to: ####
```
Just (GraphQLPostRequest {query = "query {\n  greeting(who: \"Tim\")\n}", operationName = "", variables = fromList []})
```
by a custom/temporary Aeson parser.
#### Error: ####    
:blush: I didn't record it, and now it's gone. Something along the lines of:
```
Just(Error{"query document error!definition error!query"})
```
Not exactly, but that was the gist of it.
#### Solution: ####    
Realized that the parser might be choking on the absence of the `operationName`, so tried to apply `optempty` to `nameParser` but Name was not an instance of Monoid. Changed that and ¡viola! it worked (sounds easy, but I learned something about applying Monoid to a newtype, and also picked up a prior mistake where I forgot to import Data.Text).

On a side note, the 'custom/temporary' Aeson parser does not yet solve the ambiguous `variables` problem mentioned here:
[#128](https://github.com/jml/graphql-api/pull/128)
and here:
[#135](https://github.com/jml/graphql-api/issues/135)
and obliquely here:
[#132](https://github.com/jml/graphql-api/issues/132)

### DIFF
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -25,7 +25,6 @@ import Data.Text as T (Text)
 import qualified Data.Attoparsec.Text as A
 import Test.QuickCheck (Arbitrary(..), elements, listOf)
 import Data.String (IsString(..))
-import Data.Text as T (Text, append, empty)
 
 import GraphQL.Internal.Syntax.Tokens (tok)
 
@@ -36,14 +35,6 @@ import GraphQL.Internal.Syntax.Tokens (tok)
 -- https://facebook.github.io/graphql/#sec-Names
 newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
 
--- | Allow Name to be parsed with `optempty`
---
--- Example: node = AST.Node <$> optempty nameParser
--- I.e. If nameParser fails, the Name field of AST.Node is set
--- mempty rather than propagating a failure. 
-instance Monoid Name where
-    mempty  = Name T.empty
-    mappend (Name a1) (Name a2) = Name (T.append a1 a2)
 
 -- | Create a 'Name', panicking if the given text is invalid.
 --
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -36,19 +36,14 @@ import GraphQL.Internal.Syntax.Tokens (tok)
 -- https://facebook.github.io/graphql/#sec-Names
 newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
 
+-- | Allow Name to be parsed with `optempty`
+--
+-- Example: node = AST.Node <$> optempty nameParser
+-- I.e. If nameParser fails, the Name field of AST.Node is set
+-- mempty rather than propagating a failure. 
 instance Monoid Name where
     mempty  = Name T.empty
---    mappend (Name {a}) mempty = Name {a}
---    mappend mempty (Name {b}) = Name {b}
     mappend (Name a1) (Name a2) = Name (T.append a1 a2)
---    mappend = append
---    mconcat = concat
-
---newtype Any = Any { getAny :: Bool }
-
---instance Monoid Any where
---    mempty = Any False
---    (Any b1) `mappend` (Any b2) = Any (b1 || b2)
 
 -- | Create a 'Name', panicking if the given text is invalid.
 --
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -25,6 +25,7 @@ import Data.Text as T (Text)
 import qualified Data.Attoparsec.Text as A
 import Test.QuickCheck (Arbitrary(..), elements, listOf)
 import Data.String (IsString(..))
+import Data.Text as T (Text, append, empty)
 
 import GraphQL.Internal.Syntax.Tokens (tok)
 
@@ -34,6 +35,20 @@ import GraphQL.Internal.Syntax.Tokens (tok)
 --
 -- https://facebook.github.io/graphql/#sec-Names
 newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
+
+instance Monoid Name where
+    mempty  = Name T.empty
+--    mappend (Name {a}) mempty = Name {a}
+--    mappend mempty (Name {b}) = Name {b}
+    mappend (Name a1) (Name a2) = Name (T.append a1 a2)
+--    mappend = append
+--    mconcat = concat
+
+--newtype Any = Any { getAny :: Bool }
+
+--instance Monoid Any where
+--    mempty = Any False
+--    (Any b1) `mappend` (Any b2) = Any (b1 || b2)
 
 -- | Create a 'Name', panicking if the given text is invalid.
 --
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -5,10 +5,11 @@
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 module GraphQL.Internal.Name
-  ( Name(unName)
+  ( Name(unName, Name)
   , NameError(..)
   , makeName
   , nameFromSymbol
+  , nameParser
   -- * Named things
   , HasName(..)
   -- * Unsafe functions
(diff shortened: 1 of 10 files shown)


sunwukonga:
I've created a couple of parser tests (planning validation tests) for this, but before I make a PR, I'd like some feedback on the form of the test. Importantly, the appropriateness of `optempty` putting the Name field into a state that cannot be reached through `makeName` i.e. mempty, otherwise known as "". 

```
    it "parses anonymous query documents" $ do
      let query = [r|query {
                       dog {
                         name
                       }
                     }|]
      let Right parsed = parseOnly Parser.queryDocument query
      let expected = AST.QueryDocument
                     [ AST.DefinitionOperation
                       (AST.Query
                         (AST.Node (Name mempty) [] []
                           [ AST.SelectionField
                             (AST.Field Nothing dog [] []
                               [ AST.SelectionField (AST.Field Nothing someName [] [] [])
                               ])
                           ]))
                     ]
      parsed `shouldBe` expected
```
The upside of this is that an anonymous `query` can have variables (also tested), as with:
```
 query ($userId: String!) {
  user(id: $userId) {
    id
    name(duplicateUserIdTest: $userId)
    role
    creationTime: createdAt
  }
}
```
*This query is simplified from an actual test in the test-suite of the elm-graphql client implementation.*

One other thing I've noticed is that the Node parser has a check for directives at [parser.hs:56](https://github.com/jml/graphql-api/blob/6f76aa1a37b0eb58a7c4ad556e402965619759cf/src/GraphQL/Internal/Syntax/Parser.hs#L56). Why are we checking for directives there, when the graphql [spec/documentation](http://graphql.org/learn/queries/#directives) says that they only occur on fields?

src/GraphQL/Internal/Name.hs
```diff
@@ -21,19 +21,19 @@ import Protolude
 import qualified Data.Aeson as Aeson
 import GHC.TypeLits (Symbol, KnownSymbol, symbolVal)
 import Data.Char (isDigit)
+import Data.Text as T (Text)
```
jml: Why did you do this? The `Protolude` import gets us exactly this `Text`.

src/GraphQL/Internal/Syntax/AST.hs
```diff
@@ -48,7 +48,7 @@ module GraphQL.Internal.Syntax.AST
 
 import Protolude
 
-import Data.String (IsString(..))
+--import Data.String (IsString(..))
```
jml: Please just delete, rather than comment out.

src/GraphQL/Internal/Name.hs
```diff
@@ -17,13 +18,67 @@ module GraphQL.Internal.Name
...
-  , NameError(..)
-  , unsafeMakeName
-  , makeName
-  )
+import Data.Char (isDigit)
+import Data.Text as T (Text)
+import qualified Data.Attoparsec.Text as A
+import Test.QuickCheck (Arbitrary(..), elements, listOf)
+import Data.String (IsString(..))
+import Data.Text as T (Text, append, empty)
```
jml: Don't need to import `Text` here as it's already imported four lines above.

src/GraphQL/Internal/Name.hs
```diff
@@ -17,13 +18,67 @@ module GraphQL.Internal.Name
...
+newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
+
+-- | Allow Name to be parsed with `optempty`
+--
+-- Example: node = AST.Node <$> optempty nameParser
+-- I.e. If nameParser fails, the Name field of AST.Node is set
+-- mempty rather than propagating a failure. 
+instance Monoid Name where
+    mempty  = Name T.empty
+    mappend (Name a1) (Name a2) = Name (T.append a1 a2)
```
jml: Rather than importing `T.empty` and `T.append`, I'd suggest just using `mempty` and `mempty`, e.g.

```haskell
instance Monoid Name where
  mempty = Name mempty
  mappend (Name a1) (Name a2) = Name (mappend a1 a2)
```

Looking at that, if you enable the `GeneralizedNewtypeDeriving` (sp?) language option, you could just change `Name` to derive `Monoid` to get the same definition.

src/GraphQL/Internal/Syntax/Parser.hs
```diff
@@ -51,7 +52,7 @@ operationDefinition =
   <?> "operationDefinition error!"
 
 node :: Parser AST.Node
-node = AST.Node <$> AST.nameParser
+node = AST.Node <$> optempty nameParser
```
jml: Why did you decide to go for `optempty` rather than `optional`?

My natural approach for modelling anonymous nodes would be for them to have a name of `Nothing`, rather than a name of `Name ""`.

jml:
> One other thing I've noticed is that the Node parser has a check for directives at parser.hs:56. Why are we checking for directives there, when the graphql spec/documentation says that they only occur on fields?

We're treating the [spec](http://facebook.github.io/graphql/) as authoritative. According to it, directives can be used in the following places:

* http://facebook.github.io/graphql/October2016/#sec-Language.Operations
* http://facebook.github.io/graphql/October2016/#sec-Language.Fragments
* http://facebook.github.io/graphql/October2016/#sec-Language.Fields
* http://facebook.github.io/graphql/October2016/#sec-Inline-Fragments

src/GraphQL/Internal/Syntax/Parser.hs
sunwukonga: I didn't because it was a deeper change. I would have to change the definition of
```
class HasName a where
  getName :: a -> Name
```
to
```
class HasName a where
  getName :: a -> Maybe Name
```
and then handle the Maybe everywhere that `getName` is used. Is this a direction you'd like to go?

jml: Hmm. Changing the typeclass seems wrong too, since names are only optional for nodes (IIUC). 

I guess if we went this direction, we'd drop the typeclass for nodes and re-instate `getNodeName :: Node -> Maybe Name`.

sunwukonga: Well, that's amusing. I suppose that's why Node was NOT an instance of HasName to begin with.

It would have been much simpler to change:
```
getNodeName :: Node -> Name
getNodeName (Node name _ _ _) = name
```
to
```
getNodeName :: Node -> Maybe Name
getNodeName (Node name _ _ _) = name
```
with
```
data Node = Node (Maybe Name) [VariableDefinition] [Directive] SelectionSet
             deriving (Eq,Show)
```
The only way to used `getName` of `HasName` would be to use it poorly by returning an empty `Name` when when encounter a `Node` with `Nothing` as `Name`.

jml: > I suppose that's why Node was NOT an instance of HasName to begin with.

Could well have been. To be honest, I can't remember the details. 

> It would have been much simpler…

Yeah, it would have. I'm sorry I didn't think of this at the time of #134.

src/GraphQL/Internal/Name.hs
sunwukonga: Fixed. Waiting on `Name ""` vs `Nothing` to push.

src/GraphQL/Internal/Name.hs
sunwukonga: Yes, this was right. The `GeneralizedNewtypeDeriving` with derived `Monoid` obsoleted this code.

src/GraphQL/Internal/Syntax/Parser.hs
sunwukonga: Which would you prefer to do?

1. Go back to the original `getNodeName` and have it return `Maybe Name` with the corresponding change in `Node`. Cleaner with `Maybe`, but ugly and non-unified get function.
2. Keep instance of `HasName` for `Node`, change `Node` to 
```
data Node = Node (Maybe Name) [VariableDefinition] [Directive] SelectionSet
             deriving (Eq,Show)
```
AND have the `getName` return an empty `Name` when we encounter `Nothing` instead of `Just Name "somename"`
3. Something I haven't considered...?

jml: Please:
- change `Node` to be `Node (Maybe Name) [VariableDefinition] [Directive] SelectionSet`
- go back to original `getNodeName`, except returning `Maybe Name`
- drop the `HasName` instance for `Node`

I'm sorry for the confusion here. I did a bit of digging and it looks like the spec was changed to make names for `Node` optional while we were in the middle of implementing graphql-api.

sunwukonga: Will do. Tomorrow :)

sunwukonga: Turns out to not be that simple. There is some interaction between `Name` values and the new `Maybe Name` values in Validation.hs

At line 140 of Validation.hs `visitedFrags` which contains a `Set` of `Name` and `usedFrags` which contains a `Set` of `Maybe Name` are combined. I'm not yet sure how to handle this.

src/GraphQL/Internal/Validation.hs
```diff
@@ -156,17 +156,17 @@ validate schema (AST.QueryDocument defns) = runValidator $ do
...
-    splitOps (AST.Query node@(AST.Node name _ _ _)) = Right (name, (Query, node))
-    splitOps (AST.Mutation node@(AST.Node name _ _ _)) = Right (name, (Mutation, node))
+    splitOps (AST.Query node@(AST.Node maybeName _ _ _)) = Right (maybeName, (Query, node))
+    splitOps (AST.Mutation node@(AST.Node maybeName _ _ _)) = Right (maybeName, (Mutation, node))
 
-    assertAllFragmentsUsed :: Fragments value -> Set Name -> Validation ()
+    assertAllFragmentsUsed :: Fragments value -> Set (Maybe Name) -> Validation ()
     assertAllFragmentsUsed fragments used =
-      let unused = Map.keysSet fragments `Set.difference` used
+      let unused = ( Set.map pure (Map.keysSet fragments)) `Set.difference` used
```
jml: Are names of fragments optional?

jml: They are not. What's happening here is that we've changed the type of the "observed names" set. I don't think it's strictly necessary (you can't have anonymous fragments, so there's no point recording that we've seen `Nothing`), but it only introduces a minor error message regression that we can fix easily enough later, leaving this PR a net improvement.

sunwukonga: There were two possibilities here, if I understood things correctly.
1. Leave Node Names packed in a Maybe, with the consequence that the accumulated state of StateT must change to contain "Maybe Name" as well. Since names into StateT can also come from Fragments, those names also needed to be packed into Maybes (this seemed like the safest approach to me, and although it does change error messages associated with fragments to imply that Fragment Names can be Nothing [Is this the "minor error message regression"?], in actuality, they never would be). 
2. Unpack Node Names...
    a. Under the assumption that Nodes MUST be named in this context (I was not at all sure of this) -> throw an error if Nothing.
    b. Under the assumption that Node Names can be Nothing in this context -> No way to proceed (or rather, only hacky ways to proceed, i.e. Set name to "").

jml: That's mostly right.

The `StateT` is there to track which fragments have been visited. Since all fragments have names, it's impossible for a node without a name visit a fragment. That means if we find an anonymous node, we don't need to update the `StateT`.

(Multi-tasking heavily at the moment, so I might be horribly wrong).

PR #139, opened 2017-12-22
Add tests for anonymous queries w|w/o variables

sunwukonga:
Add tests in ASTTests.hs:
- parses anonymous query documents
  - changed previous test of same name to: parses shorthand syntax documents
- parses anonymous query with variables
    
Added tests in ValidationTests.hs:
- Treats anonymous queries as valid
- Treats anonymous queries with variables as valid

### DIFF
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 module GraphQL.Internal.Name
   ( Name(unName, Name)
+  , mempty
   , NameError(..)
   , makeName
   , nameFromSymbol
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -25,7 +25,6 @@ import Data.Text as T (Text)
 import qualified Data.Attoparsec.Text as A
 import Test.QuickCheck (Arbitrary(..), elements, listOf)
 import Data.String (IsString(..))
-import Data.Text as T (Text, append, empty)
 
 import GraphQL.Internal.Syntax.Tokens (tok)
 
@@ -36,14 +35,6 @@ import GraphQL.Internal.Syntax.Tokens (tok)
 -- https://facebook.github.io/graphql/#sec-Names
 newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
 
--- | Allow Name to be parsed with `optempty`
---
--- Example: node = AST.Node <$> optempty nameParser
--- I.e. If nameParser fails, the Name field of AST.Node is set
--- mempty rather than propagating a failure. 
-instance Monoid Name where
-    mempty  = Name T.empty
-    mappend (Name a1) (Name a2) = Name (T.append a1 a2)
 
 -- | Create a 'Name', panicking if the given text is invalid.
 --
--- a/src/GraphQL/Internal/Name.hs
+++ b/src/GraphQL/Internal/Name.hs
@@ -36,19 +36,14 @@ import GraphQL.Internal.Syntax.Tokens (tok)
 -- https://facebook.github.io/graphql/#sec-Names
 newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
 
+-- | Allow Name to be parsed with `optempty`
+--
+-- Example: node = AST.Node <$> optempty nameParser
+-- I.e. If nameParser fails, the Name field of AST.Node is set
+-- mempty rather than propagating a failure. 
 instance Monoid Name where
     mempty  = Name T.empty
---    mappend (Name {a}) mempty = Name {a}
---    mappend mempty (Name {b}) = Name {b}
     mappend (Name a1) (Name a2) = Name (T.append a1 a2)
---    mappend = append
---    mconcat = concat
-
---newtype Any = Any { getAny :: Bool }
-
---instance Monoid Any where
---    mempty = Any False
---    (Any b1) `mappend` (Any b2) = Any (b1 || b2)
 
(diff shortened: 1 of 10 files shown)


jml:
Thanks for this, and indeed all your other contributions.

I’ll try to get to reviewing this as soon as I may. Please forgive the
lag—it’s the result of competing priorities, not lack of interest.

src/GraphQL/Internal/Name.hs
```diff
@@ -17,13 +19,72 @@ module GraphQL.Internal.Name
...
+import GraphQL.Internal.Syntax.Tokens (tok)
+
+-- * Name
+
+-- | A name in GraphQL.
+--
+-- https://facebook.github.io/graphql/#sec-Names
+newtype Name = Name { unName :: T.Text } deriving (Eq, Ord, Show)
+
+instance Monoid Name where
```
teh: Why is this instance useful? Would you mind adding a comment that explains why we have this instance?

src/GraphQL/Internal/Name.hs
```diff
@@ -17,13 +19,72 @@ module GraphQL.Internal.Name
...
+    mempty  = Name T.empty
+--    mappend (Name {a}) mempty = Name {a}
+--    mappend mempty (Name {b}) = Name {b}
```
teh: Are the comments here to explain the behaviour?

src/GraphQL/Internal/Name.hs
```diff
@@ -17,13 +19,72 @@ module GraphQL.Internal.Name
...
+    mappend (Name a1) (Name a2) = Name (T.append a1 a2)
+--    mappend = append
+--    mconcat = concat
+
+--newtype Any = Any { getAny :: Bool }
+
+--instance Monoid Any where
```
teh: What does this commented our block about `Any` do?

sunwukonga: Working notes, I'll remove.

src/GraphQL/Internal/Name.hs
sunwukonga: Will add comment.

src/GraphQL/Internal/Name.hs
sunwukonga: ### What it's doing
If Name is not a monoid, then `optempty` cannot operate on nameParser (because without it, it only produces a Name or a NameError). As a monoid, optempty returns a Name "SomeValidName" or Name mempty. This allows the following:
``` 
node :: Parser AST.Node
  node = AST.Node <$> optempty nameParser
                  <*> optempty variableDefinitions
                  <*> optempty directives
                  <*> selectionSet
```
This makes graphql-api`s implementation more forgiving than the spec, without any added ambiguity (because of validation that follows in our validation layer). 
### Justification & Discussion
According to the spec,  [#operation-name](http://graphql.org/learn/queries/#operation-name): 
>Up until now, we have been using a shorthand syntax where we omit both the query keyword and the query name, but in production apps it's useful to use these to make our code less ambiguous. You'll need these optional parts to a GraphQL operation if you want to execute something other than a query or pass dynamic variables.

It says, that BOTH *query keyword* AND *query name* MUST be added to *mutation*, *subscription*, or a *query* that passes *dynamic variables*. Unfortunately, it is not this strongly written however, and because *query name* (operation name) is practically useless, client implementations I've encountered don't follow it strictly.

The utility of *operation names* is in the contrived case where multiple operations are sent in a single document (_dead load because there is absolutely no use, now and possibly ever for, these extra superfluous operations in the document: look [here](https://github.com/jamesmacaulay/elm-graphql/issues/30#issuecomment-353277368) for some of the justifications I've encountered_). Note that I ultimately arguing to USE operation names in the referenced comment.

The spec is NOT strongly written enough to preclude the possibility of client implementations sending something like:
```
query {greeting(who: "Tim")}
```
which we currently fail on, but could theoretically turn into an AST.AnonymousQuery by changing:
```
operationDefinition :: Parser AST.OperationDefinition$
  operationDefinition =
        AST.Query    <$ tok "query"    <*> node
    <|> AST.Mutation <$ tok "mutation" <*> node
    <|> (AST.AnonymousQuery <$> selectionSet)
```
into something like this (untested, not 100% "query" is consumed on failure of AST.Query match):
```
operationDefinition :: Parser AST.OperationDefinition$
  operationDefinition =
        AST.Query    <$ tok "query"    <*> node
    <|> (AST.AnonymousQuery <$> selectionSet)
    <|> AST.Mutation <$ tok "mutation" <*> node
    <|> (AST.AnonymousQuery <$> selectionSet)
```
OR we can do what I have done in this PR, which is make the `Name` optional, and _have a nice day_ :)  In addition, although it's out of spec, we handle the following case as well:
 ```
query ($name : String) { greeting(who: $name) }
``` 
I.e. A request explicitly coded as a query, but without a name, and takes variables (this is not possible to misinterpret as anything other than a violation of the spec), but there is no danger in doing so because the validator will pick up ambiguity in the case of multiple operations AND no valid JSON operation name.

jml:
I've commented on #137, which I think this depends on. Will take a look at this once #137 is merged.

jml:
I've just merged #137 (thanks again for your efforts and your patience!). Can you please hit whatever buttons are necessary to give this PR an up-to-date diff from master?

sunwukonga:
I already rebased this branch on the changes I made to #137. Is that what you mean @jml ? That's why the commit [61cc96b](https://github.com/jml/graphql-api/pull/139/commits/61cc96b603115843024fdf5d1617518384b4c1f1) above failed it's tests (fixed by subsequent commit).

jml:
I just wanted to see a diff vs master. GH is currently showing changes that were made in #137 

No worries though, since I created such a diff locally. Everything looks great. Thanks for the tests!

PR #152, opened 2018-02-01
Attempt to use CircleCI 2.0

jml:
Drops documentation builds for now. We weren't really using them.

Should follow up with hlint & hindent steps.


jml:
Merging without review to give CircleCI a kick in the pants. Please feel free to do post-merge review.

Issue #150, opened 2018-01-31
Ambiguous occurence of `Type`


marcosh:
Trying to do `stack build graphql-api` in an new project, I get the following error

```
Ambiguous occurrence ‘Type’
        It could refer to either ‘Protolude.Type’,
                                 imported from ‘Protolude’ at src/GraphQL/Internal/Syntax/AST.hs:55:1-16
                                 (and originally defined in ‘GHC.Types’)
                              or ‘GraphQL.Internal.Syntax.AST.Type’,
                                 defined at src/GraphQL/Internal/Syntax/AST.hs:244:1
```

I'm quite a newbie, so I might well be wrong, but it looks like a version problem, maybe somehow related to #121

jml:
Yeah, I had to fix this locally recently. We need to change the `Protolude` import to hide `Type`.

jml:
Fixed by #154

PR #154, opened 2018-02-01
hide protolude redifined imports

marcosh:
attempt to solve #150

### DIFF
--- a/src/GraphQL/Internal/Validation.hs
+++ b/src/GraphQL/Internal/Validation.hs
@@ -58,7 +58,7 @@ module GraphQL.Internal.Validation
   , findDuplicates
   ) where
 
-import Protolude hiding ((<>))
+import Protolude hiding ((<>), throwE)
 
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NonEmpty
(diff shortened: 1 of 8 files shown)


src/GraphQL/Internal/Validation.hs
```diff
@@ -58,7 +58,7 @@ module GraphQL.Internal.Validation
   , findDuplicates
   ) where
 
-import Protolude hiding ((<>))
+import Protolude hiding ((<>), throwE)
```
teh: Do we have a `throwE` collision as well?

marcosh: it looks like it. After hiding the `Type` from `protolude`, I got the same error with this `throwE`

jml:
Thanks for tackling this, @marcosh 

This change introduces build warnings for users of protolude 0.2.0. I would rather not have the warnings, and think the right thing to do is restrict graphql-api to use protolude 0.2.1 and later.

Concretely, this means:

- [ ] Add a note to the `$next` release section of `CHANGELOG.rst` about the changed protolude requirements
- [ ] Update `package.yaml` to depend on protolude >= 0.2.1
- [ ] Update `stack.yaml` to say `protolude-0.2.1` rather than `protolude-0.2` (not 100% sure this is necessary)

@marcosh, are you OK to make these changes?

Separately, we should do a follow-up PR that breaks the build on warnings.

marcosh:
@jml I updated the PR, hope everything's fine

PR #177, opened 2018-03-06
Fix the input object example

harendra-kumar:


examples/InputObject.hs
```diff
@@ -48,7 +48,7 @@ example = interpretAnonymousQuery @Query root
 
 main :: IO ()
 main = do
-  response <- example "{ description(dogStuff: {toy: \"bone\", likesTreats: true}) }"
+  response <- example "{ description(dogStuff: {_toy: \"bone\", _likesTreats: true}) }"
```
teh: Thanks for the PR! 

I'm not entirely sure what's going on here. Is there a bug for reference? Can you maybe add a description to the commit message for why the example didn't work before and why it works with this fix?

jml: What breaks because of this?

harendra-kumar:
It generates this output without the change:

```
{"data":{"description":null},"errors":[{"message":"Could not coerce Name {unName = \"dogStuff\"} to valid value: Key not found: Name {unName = \"_toy\"}"}]}
{"data":{"description":"their favorite toy is a shoe"}}
```

With the change:

```
{"data":{"description":"likes treats and their favorite toy is a bone"}}
{"data":{"description":"their favorite toy is a shoe"}}
```

jml:
We had both approved this, and there were no outstanding comments, so I merged it.

PR #184, opened 2018-05-31
[Work in progress] Improve ast doc + adding contributions "guidelines"

theobat:
Things that I have to clarify (the list is also a WIP):
- Why is there a makefile ? 
- Why is there no default stack.yml (required for some ide tools) ? And if the stance is to consider the stack.yml as local config why is it not in .gitignore ?
- What's the role of the *TypeDefinition in the [AST](https://github.com/haskell-graphql/graphql-api/blob/master/src/GraphQL/Internal/Syntax/AST.hs#L187-L230)
- How to start the graphql-wai part and document it

Actions left:
- [ ] more doc coverage in the AST file

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -14,3 +14,5 @@ stack haddock
 
 Why is there a makefile ? 
 why is there no default stack.yml ? (required for some ide tools)
+
+-> current master branch contains code for the **next release**, current release is on hackage and its source code is available in a branch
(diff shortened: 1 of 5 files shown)


jml:
Thanks! 

Please submit the contribution guidelines update as a separate PR. Also, we prefer rST for docs, but will accept Markdown if you don't want to do rST.

theobat:
Right I'll do that, closing for now.

Issue #145, opened 2018-01-17
Handling Variables via JSON?


russmatney:
Thanks much for all the work getting this GraphQL implementation together - it's no small task.

How would you recommend using this library with variables parsed from a JSON Request? It looks like #128 would accomplish that - just want to make sure this was actually not yet supported, as the path I'm on seems to require parsing the JSON (via Aeson) prior to interpreting the query.

Would love to see that PR merged soon!

teh:
@russmatney I think for now parsing the JSON request yourself is the way to go. I haven't yet managed to convince myself that the `FromJSON` instance in #128 is what we want. My feeling is that if we have all the type information then we can do better than parsing `Enum` as a string but I may be wrong, too :)

theobat:
I think we should add an example of this in the docs as people using graphql are really used to querying their API using a separate JSON object for the variables. Can I try to make a PR for this ?

Also, I'm not sure if handling a JSON object of variables is just "delayed" until we can figure out a better implementation than #128 or whether there is no decent implementation possible given the way things are right now ? (if that makes sense)

teh:
@theobat PRs are very welcome, *especially* doc PRs!

JSON handling is delayed until someone has time to pick it up. My feeling is that the parsing should by type directed (i.e. same instance pattern matching we use for resolution) but I don't have the time in the foreseeable future to do this. It's a bit frustrating that it's stalled, but I prefer stalled over a solution I don't really believe in, especially because people can roll their own solution today (so they're not hard-blocked).

Does that make sense?

theobat:
thanks for the quick answer, it does make sense.

theobat:
So I'm playing around with this and here's what I'd like to do in another PR (But planning everything ahead seems like a safer bet): 
- First of all I believe the right way to "parse" the *variables* JSON into graphQL types is by making use of the **parsed** text query, that is the [AST.QueryDocument](https://github.com/haskell-graphql/graphql-api/blob/master/src/GraphQL/Internal/Syntax/AST.hs#L63) which is created in [parseQuery](https://github.com/haskell-graphql/graphql-api/blob/master/src/GraphQL.hs#L145), because it contains a list of [VariableDefinitions](https://github.com/haskell-graphql/graphql-api/blob/master/src/GraphQL/Internal/Syntax/AST.hs#L80) which must consist in a name and a type.
[...]

PR #186, opened 2018-06-21
Variable definitions validation

theobat:
This PR is a required step towards the resolution of #145. It improves the validation of type declaration in variable definitions. For instance it validates the fact that when a document contains $myVariable: MyVar, then the MyVar type actually exists in the schema and it's an InputType.

Remaining tasks:
- [x] test cases for the new ValidationErrors 
- [x] fix current broken EndToEnd tests

PS: this is my first "real" attempt to write haskell so please, any insight would be greatly appreciated.

**Edit: I added DefinesTypes instances on arguments because failing tests were simply not finding their argument types in the schema (schema in the sense of what is returned by makeSchema)**

### DIFF
--- a/src/GraphQL/Internal/Validation.hs
+++ b/src/GraphQL/Internal/Validation.hs
@@ -88,6 +88,7 @@ import GraphQL.Internal.Schema
   , AnnotatedType (TypeNamed, TypeNonNull)
   , NonNullType(NonNullTypeNamed)
   , getInputTypeDefinition
+  , builtinFromName
   )
 import GraphQL.Value
   ( Value
@@ -658,29 +659,34 @@ validateVariableDefinitions schema vars = do
 -- | Ensure that a variable definition is a valid one.
 validateVariableDefinition :: Schema -> AST.VariableDefinition -> Validation VariableDefinition
 validateVariableDefinition schema (AST.VariableDefinition var varType value) =
-  case validateTypeAssertion schema var varType of
-    Left e -> throwE e 
-    Right t -> VariableDefinition var t <$> traverse validateDefaultValue value
+  VariableDefinition var
+    <$> validateTypeAssertion schema var varType
+    <*> traverse validateDefaultValue value
 
 -- | Ensure that a variable has a correct type declaration given a schema.
-validateTypeAssertion :: Schema -> Variable -> AST.GType -> Either ValidationError (AnnotatedType InputType)
-validateTypeAssertion schema var varTypeAST = 
-  case typeDef of
-    Nothing -> fmap (astAnnotationToSchemaAnnotation varTypeAST) (validateVariableTypeBuiltin var varTypeNameAST) 
-    Just value -> astAnnotationToSchemaAnnotation varTypeAST . DefinedInputType <$> maybeToEither (VariableTypeIsNotInputType var varTypeNameAST) (getInputTypeDefinition value)
+validateTypeAssertion :: Schema -> Variable -> AST.GType -> Validation (AnnotatedType InputType)
+validateTypeAssertion schema var varTypeAST =
+  astAnnotationToSchemaAnnotation varTypeAST <$>
+  case lookupType schema varTypeNameAST of
+    Nothing -> validateVariableTypeBuiltin var varTypeNameAST
+    Just cleanTypeDef -> validateVariableTypeDefinition var cleanTypeDef
   where 
     varTypeNameAST = getName varTypeAST
-    typeDef = lookupType schema varTypeNameAST
 
--- | validate a variable type which has no type definition (either builtin or not in the schema)
-validateVariableTypeBuiltin :: Variable -> Name -> Either ValidationError InputType
-validateVariableTypeBuiltin var tname
-  | tname == getName GInt = Right (BuiltinInputType GInt)
-  | tname == getName GBool = Right (BuiltinInputType GBool)
-  | tname == getName GString = Right (BuiltinInputType GString)
-  | tname == getName GFloat = Right (BuiltinInputType GFloat)
-  | tname == getName GID = Right (BuiltinInputType GID)
-  | otherwise = Left (VariableTypeNotFound var tname)
+-- | Validate a variable type which has a type definition in the schema.
+validateVariableTypeDefinition :: Variable -> TypeDefinition -> Validation InputType
+validateVariableTypeDefinition var typeDef = 
+  case getInputTypeDefinition typeDef of 
+    Nothing -> throwE (VariableTypeIsNotInputType var $ getName typeDef)
+    Just value -> pure (DefinedInputType value)
+ 
+
+-- | Validate a variable type which has no type definition (either builtin or not in the schema).
+validateVariableTypeBuiltin :: Variable -> Name -> Validation InputType
+validateVariableTypeBuiltin var typeName = 
+  case builtinFromName typeName of
+    Nothing -> throwE (VariableTypeNotFound var typeName)
+    Just builtin -> pure (BuiltinInputType builtin)
 
 -- | simple translation between ast annotation types and schema annotation types
 astAnnotationToSchemaAnnotation :: AST.GType -> a -> AnnotatedType a
--- a/src/GraphQL/Internal/Validation.hs
+++ b/src/GraphQL/Internal/Validation.hs
@@ -58,6 +58,7 @@ module GraphQL.Internal.Validation
   , getResponseKey
   -- * Exported for testing
   , findDuplicates
+  , formatErrors
   ) where
 
 import Protolude hiding ((<>), throwE)
@@ -880,6 +881,11 @@ makeMap entries =
     Just dups -> throwErrors dups
 
 -- * Error handling
+
+-- | Utility function for tests, format ErrorTypes to their text representation
+-- returns a list of error messages
+formatErrors :: [ValidationError] -> [Text]
+formatErrors errors = formatError <$> errors
 
 -- | A 'Validator' is a value that can either be valid or have a non-empty
 -- list of errors.
--- a/src/GraphQL/Internal/Validation.hs
+++ b/src/GraphQL/Internal/Validation.hs
@@ -687,9 +687,9 @@ validateVariableTypeBuiltin var tname
 astAnnotationToSchemaAnnotation :: AST.GType -> a -> AnnotatedType a
 astAnnotationToSchemaAnnotation gtype schematn = 
   case gtype of
-    AST.TypeNamed asttn -> TypeNamed schematn
+    AST.TypeNamed _ -> TypeNamed schematn
     AST.TypeList (AST.ListType asttn) -> astAnnotationToSchemaAnnotation asttn schematn
-    AST.TypeNonNull (AST.NonNullTypeNamed asttn) -> TypeNonNull (NonNullTypeNamed schematn)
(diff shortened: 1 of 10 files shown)


src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+
+-- | Ensure that a variable has a correct type declaration given a schema.
+validateTypeAssertion :: Schema -> Variable -> AST.GType -> Either ValidationError (AnnotatedType InputType)
+validateTypeAssertion schema var varTypeAST = 
+  case typeDef of
+    Nothing -> fmap (astAnnotationToSchemaAnnotation varTypeAST) (validateVariableTypeBuiltin var varTypeNameAST) 
+    Just value -> astAnnotationToSchemaAnnotation varTypeAST . DefinedInputType <$> maybeToEither (VariableTypeIsNotInputType var varTypeNameAST) (getInputTypeDefinition value)
+  where 
+    varTypeNameAST = getName varTypeAST
+    typeDef = lookupType schema varTypeNameAST
```
jml: I'd inline this in the `case`, so it's obvious that we're looking at a lookup that might fail.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+validateVariableDefinition schema (AST.VariableDefinition var varType value) =
+  case validateTypeAssertion schema var varType of
+    Left e -> throwE e 
+    Right t -> VariableDefinition var t <$> traverse validateDefaultValue value
+
+-- | Ensure that a variable has a correct type declaration given a schema.
+validateTypeAssertion :: Schema -> Variable -> AST.GType -> Either ValidationError (AnnotatedType InputType)
+validateTypeAssertion schema var varTypeAST = 
+  case typeDef of
+    Nothing -> fmap (astAnnotationToSchemaAnnotation varTypeAST) (validateVariableTypeBuiltin var varTypeNameAST)
```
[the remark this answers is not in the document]
jml: Actually, since both branches of the `case` do `astAnnotationToSchemaAnnotation varTypeAST`, we should extract it out.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+validateTypeAssertion :: Schema -> Variable -> AST.GType -> Either ValidationError (AnnotatedType InputType)
+validateTypeAssertion schema var varTypeAST = 
+  case typeDef of
+    Nothing -> fmap (astAnnotationToSchemaAnnotation varTypeAST) (validateVariableTypeBuiltin var varTypeNameAST) 
+    Just value -> astAnnotationToSchemaAnnotation varTypeAST . DefinedInputType <$> maybeToEither (VariableTypeIsNotInputType var varTypeNameAST) (getInputTypeDefinition value)
+  where 
+    varTypeNameAST = getName varTypeAST
+    typeDef = lookupType schema varTypeNameAST
+
+-- | validate a variable type which has no type definition (either builtin or not in the schema)
```
jml: Nit. Please capitalize "validate" and end the sentence with a full stop.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+validateVariableTypeBuiltin :: Variable -> Name -> Either ValidationError InputType
+validateVariableTypeBuiltin var tname
```
jml: Please say `typeName` (if that's what it is).

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+  | tname == getName GInt = Right (BuiltinInputType GInt)
+  | tname == getName GBool = Right (BuiltinInputType GBool)
+  | tname == getName GString = Right (BuiltinInputType GString)
+  | tname == getName GFloat = Right (BuiltinInputType GFloat)
+  | tname == getName GID = Right (BuiltinInputType GID)
+  | otherwise = Left (VariableTypeNotFound var tname)
```
jml: I feel like there should be a better way of doing this. At least part of the answer is that there should be a function in `GraphQL.Internal.Schema` like `builtinFromName :: Name -> Maybe Builtin`.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+
+-- | simple translation between ast annotation types and schema annotation types
+astAnnotationToSchemaAnnotation :: AST.GType -> a -> AnnotatedType a
+astAnnotationToSchemaAnnotation gtype schematn = 
+  case gtype of
+    AST.TypeNamed _ -> TypeNamed schematn
+    AST.TypeList (AST.ListType asttn) -> astAnnotationToSchemaAnnotation asttn schematn
+    AST.TypeNonNull (AST.NonNullTypeNamed _) -> TypeNonNull (NonNullTypeNamed schematn)
+    AST.TypeNonNull (AST.NonNullTypeList (AST.ListType asttn)) -> astAnnotationToSchemaAnnotation asttn schematn
```
jml: This function should live in `GraphQL.Internal.Schema`.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+validateVariableTypeBuiltin :: Variable -> Name -> Either ValidationError InputType
+validateVariableTypeBuiltin var tname
+  | tname == getName GInt = Right (BuiltinInputType GInt)
+  | tname == getName GBool = Right (BuiltinInputType GBool)
+  | tname == getName GString = Right (BuiltinInputType GString)
+  | tname == getName GFloat = Right (BuiltinInputType GFloat)
+  | tname == getName GID = Right (BuiltinInputType GID)
+  | otherwise = Left (VariableTypeNotFound var tname)
+
+-- | simple translation between ast annotation types and schema annotation types
```
jml: Nit: capital first letter, capitalize "AST" and end with a full stop.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
+astAnnotationToSchemaAnnotation :: AST.GType -> a -> AnnotatedType a
+astAnnotationToSchemaAnnotation gtype schematn = 
+  case gtype of
+    AST.TypeNamed _ -> TypeNamed schematn
+    AST.TypeList (AST.ListType asttn) -> astAnnotationToSchemaAnnotation asttn schematn
```
jml: `schemaTypeName`, `astTypeName`

jml: This looks wrong. The annotated type of "list of int" should be "list of int", not "int".

src/GraphQL/Internal/Validation.hs
jml: Why does this take a `Variable`?

tests/ValidationTests.hs
```diff
@@ -27,11 +29,11 @@ someName = "name"
 
 dog :: Name
 dog = "dog"
-
+
```
jml: Nit: Please don't add unnecessary whitespace.

src/GraphQL/Internal/Validation.hs
```diff
@@ -642,16 +650,47 @@ emptyVariableDefinitions :: VariableDefinitions
...
 
 -- | Ensure that a variable definition is a valid one.
-validateVariableDefinition :: AST.VariableDefinition -> Validation VariableDefinition
-validateVariableDefinition (AST.VariableDefinition name varType value) =
-  VariableDefinition name varType <$> traverse validateDefaultValue value
+validateVariableDefinition :: Schema -> AST.VariableDefinition -> Validation VariableDefinition
+validateVariableDefinition schema (AST.VariableDefinition var varType value) =
+  case validateTypeAssertion schema var varType of
+    Left e -> throwE e 
+    Right t -> VariableDefinition var t <$> traverse validateDefaultValue value
```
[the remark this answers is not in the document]
theobat: Yes. Not sure I thoroughly grasp the entire concepts of Applicative and Monad, but as far as this example is concerned I think I understand.

src/GraphQL/Internal/Validation.hs
theobat: It takes a variable name because it enables us to give the variable name in the error message (if the given type is not a builtin type, which is the error message if we give a random type name for instance). I do believe this is valuable.

theobat:
@jml, Now I think I'm done here, for some reason the tests are marked as failed on CircleCI, but there isn't any error at any stage of the pipeline and the coverage increased everywhere:
```
#!/bin/bash -eo pipefail
STACK_YAML=stack-8.2.yaml ./scripts/hpc-ratchet

alternatives: BETTER (175 missing => 161 missing)
booleans: OK
expressions: BETTER (1494 missing => 1416 missing)
local_decls: BETTER (15 missing => 14 missing)
top_level_decls: BETTER (685 missing => 669 missing)
Exited with code 2
```

I'm still open to other changes though if things are not right yet.

jml:
@theobat Thanks, I'll take a look now. 

To fix the coverage error, you need to update the numbers in the script to match the new, better numbers: https://github.com/haskell-graphql/graphql-api/pull/186#issuecomment-401033808

tests/EndToEndTests.hs
```diff
@@ -394,6 +413,25 @@ tests = testSpec "End-to-end tests" $ do
...
+              [ "data" .= object
+                [ "dog" .= object
+                  [ "name" .= ("Mortgage" :: Text)
+                  , "doesKnowCommand" .= Null
+                  ]
+                ]
+              , "errors" .=
+                [
+                  object
+                  -- TODO: cf previous test case
```
jml: Not sure I could do this todo. Please either do it or expand on it so that someone who's not you could do it.

theobat: So this is just a reference to the one above that I did not write. I'm not sure I understand the initial remark which stated:
```
-- TODO: This error message is pretty bad. We should define
-- a typeclass for client-friendly "Show" (separate from
-- actual Show which remains extremely useful for debugging)
-- and use that when including values in error messages.
```

Now, while I understand both the idea and how to do it (I think...), it seems like a lot of work for this PR, since both error messages are pretty close, I thought it would be better to add the TODO reference to the other one, but now that I'm reading it on github, it seems rather unhelpful. I'm going to simply remove it and let the original TODO as it is

theobat:
@jml There we are ! Thanks for your reviews. As a side note, it seems like every guard containing an otherwise statement is counted as "always true" by hpc which means increasing the number of otherwise might force us to decrease coverage.. 
```
57% guards (11/19), 6 always True, 2 unevaluated
```

Next step is a PR with the actual Aeson.Object to Value translation. I won't address it before some time though.

jml:
> As a side note, it seems like every guard containing an otherwise statement is counted as "always true" by hpc

Yeah, that's something I really would like to fix in hpc (or with some kind of tooling), rather than compromising on coverage. The Haskell testing ecosystem needs to get better.

jml:
Thanks!

PR #193, opened 2018-08-07
Expand mutation and introspection support

jamesdabbs:
:wave: Hello! I've been doing some work to expand this library to support [the pi-base project](https://github.com/pi-base/server), and - while I don't think this is quite ready to go - I wanted to start a conversation about what if any of this you'd be open to merging upstream. The main things I'm looking for are:

* Support for generating a `schema.gql` - the pi-base is currently writing that to disk [here](https://github.com/pi-base/server/blob/master/src/Graph/Queries/Cache.hs#L38) so that the typescript client can [read it](https://github.com/pi-base/viewer/blob/master/package.json#L87)  and make sure the frontend stays in sync
* As part of the above, we need to understand the difference between a `query` and `mutation` document; if we're going to declare those, we might as well actually use them at dispatch time
* (Part of a previous commit, but) Object types should automatically support `__typename` introspection so that the client can cache records on `(__typename, id)`.

## TODO

I currently intend to tackle these on this branch

* [ ] Expand `Introspection` to cover as much of [the introspection spec](http://facebook.github.io/graphql/June2018/#sec-Schema-Introspection) as it reasonably can (the `__Type` > `__Field` > `__Type` part currently can't be modeled, and I don't plan on touching subscriptions or extensions yet)
* [ ] Take a closer look at `Interface` and `Union` handling (I don't use either of these in my project, so just haven't thought about them)
* [ ] Expand test coverage for new features
* [ ] General cleanup / documentation writing

Some open questions / thoughts

* I want to take a closer look at the resolver code. Importantly, fields in `queries` should probably be resolved in parallel if possible and top-level fields in `mutations` _must_ (per the spec) be resolved serially.
* I'd like for introspection to Just Work™ and not require manually mounting the exported handlers. Again, I think this'll require digging into the resolver logic more closely.


tests/EndToEndTests.hs
```diff
@@ -483,3 +509,106 @@ tests = testSpec "End-to-end tests" $ do
...
+              "dog": {
+                "name": "Mortgage"
+              },
+              "treats": 3
+            }
+          }
+        }
+      |]
+
+    -- it "can fetch the __schema" $ do
```
teh: Are the commented out tests aspirational?

jamesdabbs-procore: Yeah ... I'll fix / expand / uncomment (or remove) any of them before finalizing this for sure.

jamesdabbs-procore:
@teh - Awesome! I'll be working on finalizing and polishing this over the weekend then.

On the `__Type > __Field > __Type` thing ... see http://facebook.github.io/graphql/June2018/#sec-Schema-Introspection

The spec'd schema would need to look like

```haskell
type Type__ = Object "__Type" '[]
  '[ Field "fields" Field__
   ]

type Field__ = Object "__Field" '[]
  '[ Field "type" Type__
   ]
```
and I got the impression (from an error or comment or something; I don't remember what now) that we couldn't (currently?) support that recursive type definition. I'm hazy on that though; if the above actually does compile, then I should be able to add that part as it stands.

teh:
@jamesdabbs-procore you are right, recursion can't be expressed directly with `type`. You can work around that with an opaque (to the type system) `newtype`. E.g. from one my my projects:

```haskell
type UserQ = Object "User" '[] -- this is your GraphQL Object

-- recursion breakers:
newtype RecUserQ = RecUserQ (Handler QueryMonad UserQ)

instance HasResolver QueryMonad RecUserQ where
  type Handler QueryMonad RecUserQ = RecUserQ
  resolve (RecUserQ handler) = resolve @QueryMonad @UserQ handler

instance HasAnnotatedType RecUserQ where
  -- getAnnotatedType returns garbage but it works for now because
  -- we're not using it. See https://github.com/jml/graphql-api/issues/93
  getAnnotatedType = getAnnotatedType @Text
```

src/GraphQL.hs
```diff
@@ -135,6 +142,75 @@ interpretAnonymousQuery
...
+executeRequest (SchemaRoot qh mh) document name variables =
+  case getOperation document name variables of
+    Left e -> pure (ExecutionFailure (singleError e))
+    Right (operation, ss) -> do
+      toResult <$> case operation of
+        Query    _ _ _ -> resolve @m @queries qh (Just ss)
+        Mutation _ _ _ -> resolve @m @mutations mh (Just ss)
+
+-- | Interpret a query or mutation against a SchemaRoot
+interpretRequest
```
teh: What's the difference between `interpretRequest` and `executeRequest`? Is `interpretRequest` including the parsing of the document?

src/GraphQL/Internal/Validation.hs
```diff
@@ -926,3 +928,29 @@ instance Applicative (Validator e) where
   Validator (Left e) <*> _ = Validator (Left e)
   Validator _ <*> (Validator (Left e)) = Validator (Left e)
   Validator (Right f) <*> Validator (Right x) = Validator (Right (f x))
+
+data DefinitionType = QueryDefinition | MutationDefinition deriving (Eq, Show)
+
+getDefinitionType :: AST.QueryDocument -> Maybe Name -> Maybe DefinitionType
```
teh: Would you mind adding a comment explaining a bit more about this function does? E.g. I'm not super clear on why `DefinitionType` needs to be wrapped in a `Maybe`.

src/GraphQL/Introspection.hs
```diff
@@ -0,0 +1,317 @@
...
+
+inputObjectFieldDefinitionToAST :: InputObjectFieldDefinition -> AST.InputValueDefinition
+inputObjectFieldDefinitionToAST (InputObjectFieldDefinition name annotatedInput _) = AST.InputValueDefinition name (inputTypeToAST annotatedInput) Nothing -- FIXME
+
+typeToAST :: AnnotatedType GType -> AST.GType
+typeToAST (TypeNamed t) =
+  -- AST.TypeNamed $ AST.NamedType $ getName t
+  AST.TypeNonNull $ AST.NonNullTypeNamed $ AST.NamedType $ getName t
+typeToAST (TypeList (ListType t)) =
+  -- AST.TypeList $ AST.ListType $ AST.TypeNamed $ AST.NamedType $ getName t
```
teh: do you still need the commented out lines? If so maybe add a `TODO` saying what needs to be done to uncomment them?

teh:
@jml Not super sure why the coverage ratchet passed for 8.0 but not for 8.2. Do you have any insight / ideas?

jml:
We're not running it for 8.0—it's of little extra value for the time it would take to set up.

jml:
It's not documented very well, but here's what'll happen with the coverage ratchet:

1. tests will fail if coverage is worse
2. if you make coverage *better*, tests will also fail
3. to make tests pass, you must update `scripts/hpc-ratchet` to have the correct missing lines

The reason we fail when coverage *improves* is to make sure that step 3 actually happens. Otherwise, improvements in one PR could be undone by the next. This way, we lock in any improvements, and move closer to the goal of 100% coverage.

thiagorp:
Hey guys. Thanks for this PR. Schema introspection is something I have been looking forward to. Would you like any help here?

PR #198, opened 2018-08-25
Fix overlapping errors and improve diagnostics

dminuoso:
This fix is similar to:
https://github.com/haskell-graphql/graphql-api/pull/178

The existing code has overlapping instances, namely for:

`(l :+: r)` and `(C1 (...) :+: r)`

The second instance is unnecessary, since the instance for `C1 (...)` will produce a type error already.

Some code that triggers the overlapping instances error:

```{-# LANGUAGE DeriveGeneric    #-}

import GraphQL.Internal.API.Enum (GraphQLEnum(..))
import GHC.Generics (from, to, Generic)

data Magic = Left Int | Right deriving (Eq, Show, Generic)

instance GraphQLEnum Magic

main = main

f = enumValues @Magic```

Also an unary constructor is one that takes one argument, so I took the liberty to rename the diagnostics to nullary constructors.

A test case would be a bit more difficult because it still fails compilation.

### CODE REVIEW

**Comment by teh** [comment]
Looks good, thank you! To make the test pass can you adjust `COVERAGE_TOLERANCE` like in [this diff](https://github.com/haskell-graphql/graphql-api/pull/186/files#diff-89eb7a8abf17e1052cb2ef2731a0c695)?

**Comment by dminuoso** [comment]
@teh Done

File: src/GraphQL/Internal/API/Enum.hs [hunk 1]
```diff
@@ -86,24 +86,16 @@ instance forall conName p b. (KnownSymbol conName) => GenericEnumValues (C1 ('Me
...
-  ( TypeError ('Text "Constructor not unary: " ':<>: 'Text conName)
-  , KnownSymbol conName
-  ) => GenericEnumValues (C1 ('MetaCons conName p b) (S1 sa sb) :+: f) where
+  ) => GenericEnumValues (C1 ('MetaCons conName f s) (S1 sa sb)) where
   genericEnumValues = nonUnaryConstructorError
   genericEnumFromValue = nonUnaryConstructorError
   genericEnumToValue = nonUnaryConstructorError
 
 nonUnaryConstructorError :: a
-nonUnaryConstructorError = panic "Tried to construct enum with non-unary constructor. Should get a compile-time error instead of this."
```
jml: Please also rename this function to `nonNullaryConstructorError`. We don't have to worry about backwards compatibility as it is not exported and in `Internal` anyway.

PR #212, opened 2019-01-11
Reduce operation type repetition

thiagorp:
This PR makes the operation type as a parameter of the Operation data constructor as suggested on the comment.


thiagorp:
I am happy to be able to help a little bit 😄

teh:
I got confused by the github interface and merged master into this branch instead of the other way round. @thiagorp you OK with me cherry-picking your changes into master? Alternatively if you want to go with the PR would you mind creating a new one? There doesn't seem to be a way of backing out the merge..

thiagorp:
Hey @teh, I am ok with both options. I will create a new PR now so in case you prefer this approach you are not blocked by me. But feel free to also cherry-pick if you find it better.

PR #213, opened 2019-01-11
Use Hspec instead of Tasty

thiagorp:
Maybe there was a specific reason for using Tasty that I am unaware of and would block these changes. But the combination of `gchid` + `Tmux` + `Tasty` resulted in a very poor experience for me, I couldn't see the tests that were failing because they wouldn't show up at the bottom. `Hspec` solves this issue with a better output of the tests results.


teh:
I can't recall the exact reasons for picking tasty, but I think it might have been more usable back then? In any case no objection from me. @jml?

PR #215, opened 2019-01-30
Reduce operation type repetition

thiagorp:
View #212.


thiagorp:
I had a hard time resolving the conflict of the coverage tolerance and did a few force pushes since I'm sure I'm the only one working on this branch so far. But now everything seems to be ok.

PR #222, opened 2019-08-07
Fix issue with inability to parse mandatory types

fredefox:
Prior to this commit:

    λ parseWith (pure mempty) type_ "[B!]!"
    Done "!" (TypeList ...)

With this commit we now have:

    λ parseWith (pure mempty) type_ "[B!]!"
    Done "" (TypeNonNull ...)

The important thing to note here is that previous "[B!]!" was parsed
as a list type (not a non-null type), but all the input was not
consumed as expected.  This means that whatever parser comes after
would start by looking at the exclamation point, which belongs to the
type.

### DIFF
--- a/src/GraphQL/Internal/Syntax/Parser.hs
+++ b/src/GraphQL/Internal/Syntax/Parser.hs
@@ -212,8 +212,8 @@ directive = AST.Directive
 -- * Type Reference
 
 type_ :: Parser AST.GType
-type_ = AST.TypeList    <$> listType
-    <|> AST.TypeNonNull <$> nonNullType
+type_ = AST.TypeNonNull <$> nonNullType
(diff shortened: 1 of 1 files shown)


fredefox:
`parseWith (pure mempty)` -- BTW -- was a trick I came up with to inspect how much input attoparsec consumes.

teh:
Thanks! Can you change [this](https://github.com/haskell-graphql/graphql-api/blob/3330f898556d4956d1c785be02cc9c0ef16745f9/scripts/hpc-ratchet#L40) to the value 1364? I believe that will fix the test.

fredefox:
> Thanks! Can you change this to the value 1364? I believe that will fix the test.

Ah cool! Thanks for providing the fix, I had just glossed over it. I will update this PR later.

